PR #21, opened 2020-04-22
Set java11 artifact name for ci

skabashnyuk:
### What does this PR do?
Set java11 artifact name for ci

https://github.com/eclipse/che/issues/13180


skabashnyuk:
not needed. maven build is abandoned

PR #25, opened 2020-04-28
Setup docker image build

skabashnyuk:
### What does this PR do?
Setup docker image build

n/a

#### Release Notes
n/a

#### Docs PR
n/a


skabashnyuk:
@nickboldt @benoitf @sleshchenko wdyt about such a way to build docker image that later we can reuse for che-server docker image build?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,70 @@
...
+          ${{ runner.os }}-yarn-
+    - name: Install
+      run: yarn install --ignore-optional
+    - name: Build
+      run: yarn build -t eclipse-che-dashboardgit
+    - name: Test
+      run: yarn test
+    - name: "Docker build"
+      run: |
+        docker image prune -a -f
```
nickboldt, line 50: later you're using $(which docker). Maybe this could be pulled into a variable like `DOCKER=$(which docker)` ?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,70 @@
+#
+# Copyright (c) 2020 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+# Che-Theia workflow
```
benoitf, line 10: FYI copy paste from this line and the next one should be adapted

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,70 @@
...
+      id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
+      with:
+        path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+        key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-yarn-
+    - name: Install
+      run: yarn install --ignore-optional
+    - name: Build
+      run: yarn build -t eclipse-che-dashboardgit
```
benoitf, line 45: something looks wrong there no ?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,70 @@
...
+    - name: Test
+      run: yarn test
+    - name: "Docker build"
+      run: |
+        docker image prune -a -f
+        docker build . -t quay.io/${DOCKER_IMAGE}:latest
+    - name: "Docker Login"
+      if: "'refs/heads/master' == github.ref || startsWith(github.ref, 'refs/tags/')"
+      run:  echo ${{ secrets.QUAY_TOKEN }} |  $(which docker)  login quay.io --username skabashn --password-stdin
```
benoitf, line 54: we may provide the password directly on docker command
Also we have on other repos : QUAY_USERNAME and QUAY_PASSWORD, here username is hard coded

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+    - name: "Push Docker image (latest)"
+      if: "'refs/heads/master' == github.ref || startsWith(github.ref, 'refs/tags/')"
```
benoitf, line 57: I don't see why we check on master
as we have
```
on:
  push:
    branches:
      - master
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,70 @@
...
+      run: $(which docker) push quay.io/${DOCKER_IMAGE}:latest
+
+    - name: "Tag Docker image (versioned)"
```
benoitf, line 60: looks like I don't understand all the github.ref logic as this flow is for master branch only

Dockerfile
```diff
@@ -8,24 +8,11 @@
 # Contributors:
 #   Red Hat, Inc. - initial API and implementation
 
-# This is a Dockerfile allowing to build dashboard by using a docker container.
-# Build step: $ docker build -t eclipse-che-dashboard .
-# It builds an archive file that can be used by doing later
-#  $ docker run --rm eclipse-che-dashboard | tar -C target/ -zxf -
-FROM node:8.16.0
+FROM docker.io/httpd:2.4.43-alpine AS registry
```
benoitf, line 11: I would not remove the current Dockerfile or rename the current one

because it's two different flows: building with docker and serving the target folder

BTW we could do both build and serve on the same Dockerfile

.github/workflows/ci.yaml
skabashnyuk: sure. Missed that.

.github/workflows/ci.yaml
skabashnyuk: yep.

.github/workflows/ci.yaml
skabashnyuk: Good point. will change that.

.github/workflows/ci.yaml
skabashnyuk: that code was adopted from the one that building branches. I will polish that too. Thx.

.github/workflows/ci.yaml
skabashnyuk: that code can build tags. Not sure we need that. Idea is that to make a release you need just put a tag.

Dockerfile
skabashnyuk: That was my first iteration. Later I switch to native build because I want to cache npm modules. What do you prefer to create new Dockerfile or adopt existed? I 70/30 to create new? WDYT?

.github/workflows/ci.yaml
skabashnyuk: `$(which docker)` copy past from some sample. I don't know what is the benefit. Let replace with just `docker`

Dockerfile
benoitf: new and merge/drop once all ppl are happy

skabashnyuk:
@nickboldt @benoitf I think I fixed all your comments. Please reevaluate the code.

.github/workflows/ci.yaml
nickboldt: really, we should be using `podman` instead of `docker`, but :shrug: I bet this is another one of those "this is upstream, we shouldn't use RH stuff or we'll scare the community" arguments :D

nickboldt:
re:

```
    sed -i 's|Listen 80|Listen 8080|' /usr/local/apache2/conf/httpd.conf && \
```

... backward compliance to what?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,52 @@
...
+      id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
+      with:
+        path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+        key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-yarn-
+    - name: Install
+      run: yarn install --ignore-optional
+    - name: Build
+      run: yarn build -t eclipse-che-dashboard
```
benoitf, line 40: hi, for what is `-t eclipse-che-dashboard` ?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - name: Test
+      run: yarn test
+    - name: "Docker build"
+      run: |
+        docker image prune -a -f
+        docker build . -f Dockerfile.apache -t quay.io/${DOCKER_IMAGE}:latest
```
benoitf, line 46: hi, maybe should use another tag than latest ?
like `short-sha1 of git` or `next` ?

Dockerfile.apache
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) 2015-2018 Red Hat, Inc.
```
benoitf, line 1: just small typo: copyright 2020

Dockerfile.apache
```diff
@@ -0,0 +1,18 @@
...
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+
+FROM docker.io/httpd:2.4.43-alpine AS registry
```
benoitf, line 11: drop the `AS registry`

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,52 @@
...
+      run: yarn build
+    - name: Test
+      run: yarn test
+    - name: "Docker build"
+      run: |
+        docker image prune -a -f
+        docker build . -f Dockerfile.apache -t quay.io/${DOCKER_IMAGE}:next
+    - name: "Docker Login"
+      run:  docker login -u "${{ secrets.QUAY_USERNAME }}" -p "${{ secrets.QUAY_TOKEN }}" quay.io
+    - name: "Push Docker image (latest)"
```
benoitf, line 49:
```suggestion
    - name: "Push Docker image (next)"
```

PR #27, opened 2020-04-30
Fix parent after migrating dependencies to che repository

skabashnyuk:
### What does this PR do?
Fix parent after migrating dependencies to che repository

https://github.com/eclipse/che/issues/16780


nickboldt:
Shouldn't we just delete this?

skabashnyuk:
>Why depend on maven when we can do a docker build?

Because we not there yet. I think it would be logical to successfully migrate all processes to docker build and later delete maven files. This pr I think necessary to keep maven packaging in working state in case if we need one more release with it.

PR #29, opened 2020-05-06
Setup Apache image to be able to build che-dashboard by itself

skabashnyuk:
### What does this PR do?
Setup Apache image to be able to build che-dashboard by itself

https://github.com/eclipse/che/issues/16799


Dockerfile.apache
```diff
@@ -8,12 +8,20 @@
 # Contributors:
 #   Red Hat, Inc. - initial API and implementation
 
-FROM docker.io/httpd:2.4.43-alpine
+FROM node:8.16.2 as builder
```
benoitf, line 11:
```suggestion
FROM docker.io/node:8.16.2-alpine3.10 as builder
```

benoitf, line 11: where is build that image from CI ?

skabashnyuk: I do not understand your question. Can you please rephrase?

benoitf:
@skabashnyuk should we remove the yarn install step in gh action CI then ?
and the cache of yarn ?

Dockerfile.apache
```diff
@@ -8,12 +8,20 @@
 # Contributors:
 #   Red Hat, Inc. - initial API and implementation
 
-FROM docker.io/httpd:2.4.43-alpine
+FROM docker.io/node:8.16.2-alpine as builder
```
benoitf, line 11: it would be nice to add in ci.yaml a build of --target=builder and publish image and pull that image
```
docker pull master-image || true
```


or use buildx to benefits from layers
so

skabashnyuk:
>@skabashnyuk should we remove the yarn install step in gh action CI then ?
and the cache of yarn ?

I left it unpurpose to not download all npms during docker build. I think your idea with pull/push builder image is doing the same. Do you think it worth it?

benoitf:
I would prefer to use the docker approach as it's faster
Also if I build on my laptop I could reuse this while if cache is only at github I can't

Dockerfile.apache
```diff
@@ -8,12 +8,20 @@
 # Contributors:
```
nickboldt, line 8: Note that if you edit your Dockerfiles in VSCode or Che, calling them "apache.Dockerfile" will allow them to get syntax highlighting automatically, but calling them Dockerfile.apache means you'll have to explicitly open them with the Dockerfile editor/plugin.

(That's why we have rhel.Dockerfile instead of Dockerfile.rhel in the other repos.) :D

Dockerfile.apache
```diff
@@ -8,12 +8,20 @@
...
+FROM docker.io/httpd:2.4.43-alpine
 RUN sed -i 's|    AllowOverride None|    AllowOverride All|' /usr/local/apache2/conf/httpd.conf && \
     sed -i 's|Listen 80|Listen 8080|' /usr/local/apache2/conf/httpd.conf && \
     mkdir -p /var/www && ln -s /usr/local/apache2/htdocs /var/www/html && \
     chmod -R g+rwX /usr/local/apache2 && \
     echo "ServerName localhost" >> /usr/local/apache2/conf/httpd.conf
-COPY target/dist /usr/local/apache2/htdocs/dashboard
-RUN sed 's/<![CDATA[<base href="\/">]]/<![CDATA[<base href="\/dashboard">]]/g'  /usr/local/apache2/htdocs/dashboard/index.html
+COPY --from=builder /dashboard/target/dist/ /usr/local/apache2/htdocs/dashboard
+RUN sed -i 's/<base href="\/">/<base href="\/dashboard\/"/g'  /usr/local/apache2/htdocs/dashboard/index.html
```
nickboldt, line 27: You can make this slightly easier to read if you use the `-r` flag for extended regex. 

Then using a different field delimiter you can do:

```
RUN sed -i -r -e 's#<base href="/">#<base href="/dashboard/"#g'  /usr/local/apache2/htdocs/dashboard/index.html 
```

PR #51, opened 2020-05-26
Fix PhantomJS failure for s390x & add support for Github actions.

Prabhav-Thali:
Signed-off-by: Prabhav Thali <Prabhav.Thali1@ibm.com>

### What does this PR do?
Skips TC execution for s390x. Installation of phantomjs-prebuilt fails as phantomjs binary for s390x is not available at the provided CDN URLs. Also, we tried installing phantomjs using `apt-get install phantomjs` but phantomjs crashes while executing `yarn test`. So currently skipping `yarn test` for s390x.

Also, added s390x support in GitHub Actions.

This PR is part of [this ](https://github.com/eclipse/che/issues/16655)initiative and also fixes phantomjs failure for s390x mentioned in [#16983](https://github.com/eclipse/che/issues/16983).


akurinnoy:
Hello @Prabhav-Thali , thanks for the PR!

Have you considered not to install phantomjs and other dev dependecies for `linux/s390x` platform at all? I mean just run `yarn install --production`.

Prabhav-Thali:
@akurinnoy Thanks for the suggestion. I tried using `yarn install --production` in case of s390x but it produced multiple errors. However, that command passed and further the build failed on `yarn build` command as module `wrench` was not found. After installing wrench, build failed for some other missing module and same scenario was observed for multiple dev dependencies. PFB build logs for the same.

[build_s390x.log](https://github.com/eclipse/che-dashboard/files/4687503/build_s390x.log)

Prabhav-Thali:
@akurinnoy  Any update on the merge?

apache.Dockerfile
```diff
@@ -10,12 +10,16 @@
 
 FROM docker.io/node:8.16.2 as builder
 
+RUN if [ "$(uname -m)" = "s390x" ]; then apt-get update; apt-get install -y phantomjs; fi
```
benoitf, line 13: hello, it looks like too hardcoded for a specific arch

maybe you need to provide a DOCKER_ARG to have skip-tests option and ignore optional

Prabhav-Thali: @benoitf If my understanding is correct, passing DOCKER_ARG can impact Github Actions and will need to be taken care of with additional `docker build buildx` commands. Please let me know your thoughts.

akurinnoy: @benoitf are you OK to merge this PR as is? or do you want changes?

PR #64, opened 2020-07-01
Better error handling

akurinnoy:
### What does this PR do?

This PR improves error handling so now SSO-related errors will look similarly to the following screenshot:

Screenshot 2020-06-30 at 15 51 05

fixes https://github.com/eclipse/che/issues/17154

### DIFF
--- a/src/app/index.module.ts
+++ b/src/app/index.module.ts
@@ -70,34 +70,40 @@ function buildKeycloakConfig(keycloakSettings: any): any {
 interface IResolveFn<T> {
   (value?: T | PromiseLike<T>): void;
 }
-
-interface IRejectFn<T> {
-  (reason?: any): void;
+interface IRejectFn {
+  (reason: string): void;
 }
 
 function keycloakLoad(keycloakSettings: any) {
-  return new Promise((resolve: IResolveFn<any>, reject: IRejectFn<any>) => {
+  return new Promise((resolve: IResolveFn<any>, reject: IRejectFn) => {
     const script = document.createElement('script');
     script.async = true;
     script.src = keycloakSettings['che.keycloak.js_adapter_url'];
     script.addEventListener('load', resolve);
     script.addEventListener('error', () => {
       return cheBranding.ready.then(() => {
-        reject(`<div class="header"><i class="fa fa-warning"></i><p>Certificate Error</p></div>
-   <div class="body"><p>Your ${cheBranding.getProductName()} server may be using a self-signed certificate.
-     To resolve this issue, try to import the servers CA certificate into your browser, as described
-   <a href="${cheBranding.getDocs().certificate}" target="_blank">here</a>.</p>
-   <p>After importing the certificate, refresh your browser.</p>
-   <p><a href="/">Refresh Now</a></p></div>`);
+        console.error('Keycloak adapter loading error.');
+        reject(`
+          <div class="header"><i class="fa fa-warning"></i><p>Certificate Error</p></div>
+          <div class="body">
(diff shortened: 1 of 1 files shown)


src/app/index.module.ts
```diff
@@ -219,12 +247,12 @@ angular.element(document).ready(() => {
     cheBranding.ready.then(() => {
       (angular as any).resumeBootstrap();
     });
-  }).catch((error: string) => {
-    console.error(`Can't GET "/api". ${error ? 'Error: ' : ''}`, error);
-    if (!hasCertificateError) {
```
sleshchenko, line 224: Do we still how TLS certificate issue error message?

src/app/index.module.ts
```diff
@@ -193,23 +225,19 @@ angular.element(document).ready(() => {
         redirectUrl: keycloakSettings['che.keycloak.redirect_url.dashboard']
       };
       return keycloakInit(keycloakAuth.config, initOptions);
-    }).catch((error: any) => {
-      if (keycloakSettings['che.keycloak.js_adapter_url']) {
-        hasCertificateError = true;
-      }
-      return Promise.reject(error);
+    }).catch((errorMessage: string) => {
+      return Promise.reject(errorMessage);
```
olexii4, line 229: Do we really need it
```
.catch((errorMessage: string) => {
      return Promise.reject(errorMessage);
})
```
?
Maybe leave as is(without catching at all).
WDYT?

src/app/index.module.ts
```diff
@@ -219,12 +247,12 @@ angular.element(document).ready(() => {
...
     });
-  }).catch((error: string) => {
-    console.error(`Can't GET "/api". ${error ? 'Error: ' : ''}`, error);
-    if (!hasCertificateError) {
-      error = `${error}<div>Click <a href="/">here</a> to reload page.</div>`
+  }).catch((errorMessage: string) => {
+    if (hasSSO) {
+      showErrorMessage(errorMessage);
+    } else {
+      showErrorMessage(`${errorMessage}<div>Click <a href="/">here</a> to reload page.</div>`);
```
olexii4, line 254: Why do we have a difference in this case?
```
...<div>Click <a href="/">here</a> to reload page.</div>
```

src/app/index.module.ts
```diff
@@ -174,11 +196,21 @@ const keycloakAuth = {
...
+  const promise = fetch('/api/keycloak/settings')
+    .then(response => response.json())
+    .catch(error => {
+      const errorMessage = `Can't get keycloak settings: ` + error;
+      console.error(errorMessage);
+      return Promise.reject(errorMessage);
+    });
+  let hasSSO = false;
   promise.then((keycloakSettings: any) => {
+    if (keycloakSettings['che.keycloak.js_adapter_url']) {
```
olexii4, line 208:
```
if (!keycloakSettings['che.keycloak.js_adapter_url']) {
  return;
}

hasSSO = true;
```

WDYT?

src/app/index.module.ts
akurinnoy: yes, makes sense, thanks

src/app/index.module.ts
akurinnoy: I slightly refactored the code you'd added to handle errors related to certificate issues. I assume that's because in case of issue related to SSO/certificate user need to somehow resolve the reason of the issue, in case of other issues simple page refreshing may help.

src/app/index.module.ts
akurinnoy: correct

PR #84, opened 2020-08-31
Storage type improvements

akurinnoy:
### What does this PR do?

PatternFly switch component:
- added attribute to disable the component

Storage Type switch on Get Started tab:
- correctly initialize the toggle with the default storage type
- disable the toggle in case when the only storage type is available
Screenshot_20200901_131000

Storage Type select on Custom Workspace tab:
- correctly initialize storage type select with the default storage type value
- disable the select in case when the only storage type is available
Screenshot_20200901_131506

- fixes updating the devfile editor on selecting the storage type

fixes https://github.com/eclipse/che/issues/17725

### DIFF
--- a/src/app/get-started/custom-workspace-tab/storage-type-row/storage-type-row.controller.ts
+++ b/src/app/get-started/custom-workspace-tab/storage-type-row/storage-type-row.controller.ts
@@ -28,7 +28,7 @@ export class StorageTypeRowController implements ng.IController, IStorageTypeRow
   ];
 
   // component bindings
-  isReady: boolean = false;
+  storageType?: StorageType;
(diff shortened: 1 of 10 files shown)


sleshchenko (commented):
At this point there is still bug exists: workspace is created by get started page while ephemeral mode is enabled by default. But no more bugs found.

akurinnoy:
@sleshchenko I hope that bug's fixed now, please test one more time

ibuziuk:
thanks folks, any chance we can backport it to 7.18.1 also?

akurinnoy:
@ibuziuk I'll cherry-pick changes to that branch as well

PR #97, opened 2020-10-01
UD customization improvement

akurinnoy:
### What does this PR do?

This PR improves UD customization abilities in following way: 
- get rid of branding data duplication,
- handle `product.json` loading error (see screenshots), and
- `product.json` is empty by default and will contain only patches to override default branding values

## Screenshots

<details>
<summary>Mailformed JSON file</summary>

Screenshot 2020-10-01 at 14 19 34

</details>

<details>
<summary>`product.json` is not available</summary>

Screenshot 2020-10-01 at 14 18 28

</details>

fixes https://github.com/eclipse/che/issues/17741


sleshchenko (commented):
I think product.json unavailability is mainly the issue of packaging, so user is not able to solve it with just refreshing. So, `Try Shift+Refresh or contact cluster admin if it did not help` should be a bit better

akurinnoy:
@sleshchenko updated message:

Screenshot 2020-10-02 at 18 19 27

.gitignore
```diff
@@ -1,11 +1,10 @@
-src/app/colors/che-output-colors.constant.js
 src/app/colors/che-output-colors.constant.ts
-src/app/colors/codenvy-color.constant.js
 src/app/colors/codenvy-color.constant.ts
 src/app/colors/template/
-src/app/proxy/proxy-settings.constant.js
 src/app/proxy/proxy-settings.constant.ts
 src/app/proxy/template/
+# src/components/branding/branding.constant.ts
```
olexii4, line 6: Could you remove it?

sleshchenko (approved):
Nice improvement to avoid unneeded duplicates.

Please as described in the issue:
- create a PR in che-docs to fix the link to place where links are configured;
- create a PR in rh-che to avoid specifying values which duplicate upstream values;

.gitignore
akurinnoy: I've uncommented that line because `branding.constant.ts` should be ignored as generated on project build.

src/app/keycloak-setup.ts
```diff
@@ -0,0 +1,204 @@
...
+  // initialize keycloak
+  let keycloak;
+  try {
+    await loadKeycloakAdapter(keycloakSettings, cheBranding)
+    keycloak = await initKeycloak(keycloakAuth.config, keycloakSettings);
+
+    keycloakAuth.isPresent = true;
+    keycloakAuth.keycloak = keycloak;
+    /* tslint:disable */
+    window['_keycloak'] = keycloak;
```
olexii4, line 60: Why do we need this object ```window['_keycloak'] = keycloak;``` when we already have one ```keycloakAuth.keycloak = keycloak;```?

olexii4: I think we should use only one.

akurinnoy: good catch! it's not used after this refactoring, I'll remove it.

src/app/keycloak-setup.ts
```diff
@@ -0,0 +1,204 @@
...
+      throw new Error(`${e}<div>Click <a href="/">here</a> to reload page.</div>`);
+    }
+  }
+
+  return keycloakAuth;
+}
+
+function buildKeycloakConfig(keycloakSettings: any): any {
+  const theOidcProvider = keycloakSettings['che.keycloak.oidc_provider'];
+  if (!theOidcProvider) {
```
olexii4, line 84: We can simplify this code:
```
if (theOidcProvider) {
    return {
      oidcProvider: theOidcProvider,
      clientId: keycloakSettings['che.keycloak.client_id']
    };
}

return {
      url: keycloakSettings['che.keycloak.auth_server_url'],
      realm: keycloakSettings['che.keycloak.realm'],
      clientId: keycloakSettings['che.keycloak.client_id']
};
```

olexii4:
```
enum KeycloakSettings {
  url = 'che.keycloak.auth_server_url',
  realm = 'che.keycloak.realm',
  clientId ='che.keycloak.client_id'
}
```
WDYT?

.gitignore
olexii4: It is a result file of the task 'branding'. We shouldn't commit it.

src/components/branding/che-branding.ts
```diff
@@ -50,12 +54,19 @@ export class CheBranding {
    * Update branding data.
    */
   private updateData(): Promise<void> {
-    return angular.element.get(`${ASSET_PREFIX}product.json`).then(branding => {
+    return angular.element.getJSON(`${ASSET_PREFIX}product.json`).then(branding => {
```
olexii4, line 57: It does not have a sense after the gulp task 'branding'.

olexii4: With these changes, we generate a file with constant during each build.
Should we request the same content in additional?

akurinnoy: Not sure I've got what you mean, could you elaborate more on this?

src/components/branding/branding.constant.ts
```diff
@@ -11,123 +11,62 @@
...
-  }
-}
-
-export enum TogglableFeature {
-  WORKSPACE_SHARING = 'workspaceSharing',
-  KUBERNETES_NAMESPACE_SELECTOR = 'kubernetesNamespaceSelector',
-}
+/**
+ * Generated content. Any changes will be lost after project build.
+ */
```
olexii4, line 16: We shouldn't commit this file. It is 'Generated content'.

akurinnoy: We've already have such files committed into repository, do you think we should treat only this file differently, or should we get rid of all generated files?

src/components/branding/che-branding.ts
olexii4: We should decide what to use. We can generate and use the file with constants or we should request a 'product.json' file. And we have a modal error message without branding.

I prefer the next idea - all branding resources move to a separate container.

src/components/branding/branding.constant.ts
olexii4: I am sure we should rid of all files which are generated each build.

src/components/branding/che-branding.ts
akurinnoy: Correct me if I'm wrong, but in our previous discussions we've decided:

1. get rid of branding data duplication in the dashboard project
2. `product.json` should contain only patch to overwrite default branding.
3. show an error page in case if `product.json` is not available (doesn't matter empty or not)

And there was nothing about not to request the `product.json`.

Please read also the description to [this issue](https://github.com/eclipse/che/issues/17741).

src/components/branding/branding.constant.ts
akurinnoy: I've just noticed that this makes the IDE complaining about wrong imports of those files. So I'm reverting the commit that removes this file.

src/app/keycloak-setup.ts
akurinnoy: done

src/app/keycloak-setup.ts
akurinnoy: done

src/components/branding/che-branding.ts
sleshchenko: I confirm - we're going to merge two sources - one from source-code + product.json that is going to be mandatory requested, and is supposed to contain patch from downstream. Dashboard is failed to load if product.json is not available.

PR #101, opened 2020-10-12
Do not fail in Single-User Mode is keycloak is not found

sleshchenko:
### What does this PR do?
Makes dashboard don't fail if api/keycloak/settings is not found

It fixes https://github.com/eclipse/che/issues/18095
I've tested multiuser and single-user run on minikube with deployed Helm chart


olexii4 (approved):

sleshchenko:
@olexii4 thanks for providing fix for dev mode. Retested after your change with helm chart and docker image - works just fine.

PR #110, opened 2020-11-30
Upgrade typescript plugin to latest

sleshchenko:
### What does this PR do?
Upgraded typescript plugin to 1.35.1

Fixes https://github.com/eclipse/che/issues/18478

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/devfile.yaml
+++ b/devfile.yaml
@@ -4,7 +4,7 @@ metadata:
 components:
   - alias: typescript-plugin
     type: chePlugin
-    id: che-incubator/typescript/1.30.2
+    id: che-incubator/typescript/latest
   - alias: dash-dev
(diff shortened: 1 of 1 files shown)


akurinnoy (approved):

devfile.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
 components:
   - alias: typescript-plugin
     type: chePlugin
-    id: che-incubator/typescript/1.30.2
+    id: che-incubator/typescript/1.35.1
```
benoitf, line 7: I think you should use latest as only one version is kept by the plugin-registry

sleshchenko: Fair point.

PR #119, opened 2021-01-05
Multiarch build support by docker buildx

Prabhav-Thali:
### What does this PR do?
Enables docker buildx support in GitHub Actions release job to build and publish multiarch docker image.

This refers to [#17124](https://github.com/eclipse/che/issues/17124).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,30 +7,110 @@ on:
         description: 'The version that is going to be released. Should be in format 7.y.z'
         required: true
         default: '7.y.z'
+
+env:
+  IMAGE: quay.io/eclipse/che-dashboard
+
 jobs:
-  build:
-    name: Create Release
-    runs-on: ubuntu-latest
+
+  build-images:
+    runs-on: ubuntu-18.04
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: [amd64,arm64,ppc64le,s390x]
+    outputs:
+      amd64: ${{ steps.result.outputs.amd64 }}
+      arm64: ${{ steps.result.outputs.arm64 }}
+      ppc64le: ${{ steps.result.outputs.ppc64le }}
+      s390x: ${{ steps.result.outputs.s390x }}
     steps:
-      - uses: actions/checkout@v2
-      - name: Login to docker.io
+      -
+        name: "Checkout Che Dashboard source code"
+        uses: actions/checkout@v2
+      -
+        name: "Set up QEMU"
+        uses: docker/setup-qemu-action@v1
+      -
+        name: "Set up Docker Buildx ${{ matrix.arch }}"
+        uses: docker/setup-buildx-action@v1
+      -
+        name: "Docker quay.io Login"
         uses: docker/login-action@v1
         with:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_PASSWORD }}
-          registry: docker.io
-      - name: Login to quay.io
-        uses: docker/login-action@v1
-        with:
+          registry: quay.io
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_PASSWORD }}
+      -
+        name: "Build and push ${{ matrix.arch }}"
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./apache.Dockerfile
+          platforms: linux/${{ matrix.arch }}
+          push: true
+          tags: ${{ env.IMAGE }}:${{ github.event.inputs.version }}-${{ matrix.arch }}
+      -
+        id: result
+        name: "Build result outputs version"
+        if: ${{ success() }}
+        run: echo "::set-output name=${{ matrix.arch }}::${{ github.event.inputs.version }}-${{ matrix.arch }}"
+
+  create-manifest:
+    needs: build-images
+    runs-on: ubuntu-18.04
+    steps:
+      -
+        name: "Docker quay.io Login"
+        uses: docker/login-action@v1
+        with:
           registry: quay.io
-      - name: Create Release
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
+      -
+        name: "Create and push manifest"
+        run: |
+          AMD64_VERSION="${{ needs['build-images'].outputs.amd64 }}"
+          ARM64_VERSION="${{ needs['build-images'].outputs.arm64 }}"
+          PPC64LE_VERSION="${{ needs['build-images'].outputs.ppc64le }}"
+          S390X_VERSION="${{ needs['build-images'].outputs.s390x }}"
+
+          if [[ -z "$AMD64_VERSION" || \
+               -z "$ARM64_VERSION" || \
+               -z "$PPC64LE_VERSION" || \
+               -z "$S390X_VERSION" ]]; then
+            echo "[!] The job 'build-images' fails on some of the architectures. Can't create complete manifest.";
+            exit 1;
+          fi
+
+          AMEND=""
+          AMEND+=" --amend ${{ env.IMAGE }}:$AMD64_VERSION";
+          AMEND+=" --amend ${{ env.IMAGE }}:$ARM64_VERSION";
+          AMEND+=" --amend ${{ env.IMAGE }}:$PPC64LE_VERSION";
+          AMEND+=" --amend ${{ env.IMAGE }}:$S390X_VERSION";
+
+          docker manifest create ${{ env.IMAGE }}:${{ github.event.inputs.version }} $AMEND
+          docker manifest push ${{ env.IMAGE }}:${{ github.event.inputs.version }}
+      -
+        id: result
+        name: "Manifest result"
+        if: ${{ success() }}
+        run: echo "Manifest was created and pushed successfully"
+
+  tag-release:
+    needs: create-manifest
+    runs-on: ubuntu-18.04
+    steps:
+      -
+        name: "Checkout Che Dashboard source code"
+        uses: actions/checkout@v2
+      -
+        name: "Tag release"
         run: |
           git config --global user.name "Mykhailo Kuznietsov"
           git config --global user.email "mkuznets@redhat.com"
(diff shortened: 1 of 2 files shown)


akurinnoy:
@mkuznyetsov please have a look

akurinnoy:
Hello @Prabhav-Thali,
we used to have multiplatform docker buildx build in our GH workflow but it happened to be rather unstable. Instead, we build images for each platform in a separate builder and then create and push the manifest. This approach seems to be stable enough; you can see it in [ci-multiarch.yaml](https://github.com/eclipse/che-dashboard/blob/master/.github/workflows/ci-multiarch.yaml). 

Can you please align this PR with the approach from `ci-multiarch.yaml`?

make-release.sh
```diff
@@ -108,8 +108,7 @@ fi
 if [[ $TRIGGER_RELEASE -eq 1 ]]; then	
   # push new branch to release branch to trigger CI build	
   git checkout "${BRANCH}"	
-  docker build -t "${QUAY_REPO}" -f apache.Dockerfile .	
-  docker push "${QUAY_REPO}"	
+  docker buildx build --platform linux/amd64,linux/s390x -t "${QUAY_REPO}" -f apache.Dockerfile  --push .
```
nickboldt, line 111: please use a PLATFORMS file like in this PR https://github.com/eclipse/che-devfile-registry/pull/336

nickboldt: also note that s390x fails w/ qemu and ubuntu, so this isn't a viable solution. Might have to look at travis + IBM services.

Prabhav-Thali:
@akurinnoy yes sure, I'll align this PR with the approach from ci-multiarch.yaml

Prabhav-Thali:
Hi @akurinnoy,
In `ci-multiarch.yaml` approach, I have observed that [ci.yml](https://github.com/eclipse/che-dashboard/blob/master/.github/workflows/ci.yaml) is kept as is and a new file i.e [ci-multiarch.yml](https://github.com/eclipse/che-dashboard/blob/master/.github/workflows/ci-multiarch.yaml) is created. Will this be the same in case of release workflow?

nickboldt:
> Hi @akurinnoy,
> In `ci-multiarch.yaml` approach, I have observed that [ci.yml](https://github.com/eclipse/che-dashboard/blob/master/.github/workflows/ci.yaml) is kept as is and a new file i.e [ci-multiarch.yml](https://github.com/eclipse/che-dashboard/blob/master/.github/workflows/ci-multiarch.yaml) is created. Will this be the same in case of release workflow?

See discussion in https://github.com/eclipse/che-dashboard/pull/177

akurinnoy:
Hi @Prabhav-Thali,
We're going to have the multiarch build as default and disable the single-arch build at all (https://github.com/eclipse/che-dashboard/pull/177). So I think that providing necessary changes to the existing release workflow will be the best option.

Prabhav-Thali:
@akurinnoy I have updated the PR with necessary changes. Please review.

akurinnoy (approved):
LGTM

@nickboldt please review

.github/workflows/release.yml
```diff
@@ -7,27 +7,119 @@ on:
...
+          registry: docker.io
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
-          registry: docker.io
-      - name: Login to quay.io
+      -
+        name: "Build and push ${{ matrix.arch }}"
+        uses: docker/build-push-action@v2
+        with:
+          context: .
```
nickboldt, line 56: Are you missing a cache step here?

https://github.com/eclipse/che-dashboard/blob/master/.github/workflows/ci-multiarch.yaml#L65-L66

.github/workflows/release.yml
```diff
@@ -7,27 +7,119 @@ on:
         description: 'The version that is going to be released. Should be in format 7.y.z'
         required: true
         default: '7.y.z'
+
+env:
+  IMAGE: quay.io/eclipse/che-dashboard
+
```
nickboldt, line 13: are you missing a cache image here? 

https://github.com/eclipse/che-dashboard/blob/master/.github/workflows/ci-multiarch.yaml#L21

.github/workflows/release.yml
```diff
@@ -7,27 +7,119 @@ on:
...
+        if: ${{ success() }}
+        run: echo "Manifest was created and pushed successfully"
+
+  create-release:
+    if: always()
+    needs: create-manifest
+    runs-on: ubuntu-18.04
+    steps:
+      -
+        name: "Checkout Che Dashboard source code"
```
nickboldt, line 117: I'm sure it doesn't impact much but since you already checked out code on line 30 https://github.com/eclipse/che-dashboard/pull/119/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R30 do you need to do it again?

.github/workflows/release.yml
```diff
@@ -7,27 +7,119 @@ on:
...
+            exit 1;
+          fi
+          docker manifest create ${{ env.IMAGE }}:${{ github.event.inputs.version }} $AMEND
+          docker manifest push ${{ env.IMAGE }}:${{ github.event.inputs.version }}
+      -
+        id: result
+        name: "Manifest result"
+        if: ${{ success() }}
+        run: echo "Manifest was created and pushed successfully"
+
```
nickboldt, line 110: would a more complete test be to check if the arches we expect to have built `[amd64,arm64,ppc64le,s390x]` exist in the manifest?

```
$➔ echo -n "["; skopeo inspect docker://quay.io/eclipse/che-dashboard:multiarch-next --raw | jq -r '.manifests[].platform.architecture' | xargs | tr " " "," | sed -r -e "s#\$#\]#"
[amd64,arm64,ppc64le,s390x]
```

.github/workflows/release.yml
Prabhav-Thali: As I have added create-release as a separate job, we need to checkout the code again.

.github/workflows/release.yml
```diff
@@ -7,27 +7,119 @@ on:
...
+  create-release:
+    if: always()
+    needs: create-manifest
+    runs-on: ubuntu-18.04
+    steps:
+      -
+        name: "Checkout Che Dashboard source code"
+        uses: actions/checkout@v2
+      -
+        name: "Create release"
```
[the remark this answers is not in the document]
Prabhav-Thali: sure, we can name it as Tag release instead of Create release.

.github/workflows/release.yml
Prabhav-Thali: @nickboldt Actually, I have not observed caching for release workflow in other packages so had not included it here. But if we need to add caching then will it use the same image as used in [ci-multiarch.yaml](https://github.com/eclipse/che-dashboard/blob/master/.github/workflows/ci-multiarch.yaml) i.e docker.io/cheincubator/che-dashboard:cache ?

nickboldt: yeah, I suppose so... unless that will cause collisions with the nightly builds? @sleshchenko  your thoughts?

sleshchenko: Well, I'm not sure but I think caching in the CI-multiarch is broken since all archs which are run in parallel use the same tag for cache
```
> docker pull  docker.io/cheincubator/che-dashboard:cache
cache: Pulling from cheincubator/che-dashboard
no matching manifest for linux/amd64 in the manifest list entries
```

As about caching for release, I assume it's redundant, the only stuff can be cached - dependencies, which depends on 
```
COPY package.json /dashboard/
COPY yarn.lock /dashboard/
```
But since package.json has [specified version](https://github.com/eclipse/che-dashboard/blob/master/package.json#L3), which is different from release to release, and from next builds, the cache won't be used.

The answer for cache here is: 
- how many times we trigger release build for the same version?
I think it must be 1)

.github/workflows/release.yml
sleshchenko: the cache is under discussion in https://github.com/eclipse/che-dashboard/pull/119#discussion_r590943118

.github/workflows/release.yml
```diff
@@ -7,27 +7,119 @@ on:
...
+        uses: docker/setup-buildx-action@v1
+      -
+        name: "Docker quay.io Login"
         uses: docker/login-action@v1
         with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
+      -
+        name: "Docker docker.io Login"
```
sleshchenko, line 46: Seems we don't need to login to docker.io here anymore

.github/workflows/release.yml
sleshchenko: amdI would say we even should not push manifest if any of expected arch are not built. See https://github.com/eclipse/che-dashboard/pull/177#discussion_r590338827

sleshchenko (approved):
LGTM

The changes I expect to be made before merging:
1. Remove redundant login to docker.io
2. Release should not pass if any of arch builds failed.

sleshchenko (approved):

Prabhav-Thali:
@sleshchenko @akurinnoy I'll be pushing one more minor change. Currently, testing is in progress.

.github/workflows/release.yml
akurinnoy: I think it would even better not to start this job if the previous job fails to build an image of some architecture.

Prabhav-Thali: I have removed `continue-on-error: true` such that in case of failure on any of the architectures, next job won't be executed.

akurinnoy (approved):

.github/workflows/release.yml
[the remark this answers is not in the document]
sleshchenko: @Prabhav-Thali it seems to left unaddressed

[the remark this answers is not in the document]
Prabhav-Thali: @sleshchenko my bad, updated it on line 100 but missed this one. Have updated now

olexii4 (approved):

sleshchenko (approved):
Good job is done in the PR, I can't ask for more :+1: 

Any other minor fixes or improvements can be done in further PR.
Dear Reviewers, if you have any suggestions/comments please leave them on merged PR.

PR #136, opened 2021-01-20
update GH workflows

akurinnoy:
### What does this PR do?

This PR improves GH workflows in order to:

- use GH actions to build and push images;
- build multi-platform images using isolated builders;
- build test image of che-server if che-dashboard image is built successfully for at least the default platform.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   "homepage": "https://github.com/che-incubator/che-dashboard-next#readme",
   "engines": {
     "yarn": ">=1.20.0",
-    "node": ">=10.20.1"
+    "node": ">=12"
   },
   "scripts": {
     "compile": "webpack --config webpack.config.prod.js",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -18,7 +18,7 @@
   "homepage": "https://github.com/che-incubator/che-dashboard-next#readme",
   "engines": {
     "yarn": ">=1.20.0",
-    "node": ">=10.20.1"
+    "node": ">=12"
```
olexii4, line 21: @akurinnoy You are going to change the minimum requirement could you fix ```devfile.yaml``` too:
https://github.com/eclipse/che-dashboard/blob/main-react/devfile.yaml#L22
```docker.io/node:10.22.1``` -> ``` docker.io/node:12 ```

benoitf (approved):

package.json
akurinnoy: rebased, updated.

olexii4 (approved):

PR #148, opened 2021-02-01
Update dockerfile for multiplatform build

akurinnoy:
### What does this PR do?

This PR updates dockerfile in order to fix multiplatform build.

N/a

### DIFF
--- a/apache.Dockerfile
+++ b/apache.Dockerfile
@@ -7,7 +7,8 @@
 #
 # Contributors:
 #   Red Hat, Inc. - initial API and implementation
-FROM docker.io/node:12.20.1 as builder
+
+FROM docker.io/node:12.20.1-alpine3.12 as builder
 
 COPY package.json /dashboard/
 COPY yarn.lock /dashboard/
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        platform: [linux/amd64, linux/arm64v8, linux/ppc64le, linux/s390x]
+        platform: [linux/amd64, linux/arm64, linux/ppc64le, linux/s390x]
     env:
       IMAGE_FULL: quay.io/eclipse/che-dashboard:next
       CACHE_IMAGE_FULL: docker.io/cheincubator/che-dashboard:cache
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -81,7 +81,7 @@ jobs:
         platform: [linux/amd64]
         default: [true]
         include:
-          - platform: linux/arm64v8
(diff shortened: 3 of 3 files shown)


akurinnoy:
[dashboard-next-ci-test]

apache.Dockerfile
```diff
@@ -16,7 +17,7 @@ RUN yarn --network-timeout 600000 && yarn install
 COPY . /dashboard/
 RUN yarn compile
 
-FROM docker.io/httpd:2.4.43-alpine
+FROM docker.io/$TARGETARCH/httpd:2.4.43-alpine
```
benoitf, line 20: why do we need targetarch ?

apache.Dockerfile
```diff
@@ -7,7 +7,8 @@
 #
 # Contributors:
 #   Red Hat, Inc. - initial API and implementation
-FROM docker.io/node:12.20.1 as builder
+
+FROM docker.io/$TARGETARCH/node:12.20.1-buster as builder
```
benoitf, line 11: Hello,
I'm not sure to understand  why we need targetarch there and buster ?

These are multi arch images already

akurinnoy: hello @benoitf ,

if I understand it correctly, node image `node:12.20.1` doesn't support architectures `linux/s390x` and `linux/ppc64le`. But node image `node:12.20.1-buster` supports all architectures we want to target.

Another issue here that I got following error for `linux/arm64v8` when I build image from `buster` without specifying `$TARGETARCH`:
```
...
 => ERROR [linux/arm64v8 internal] load metadata for docker.io/library/node:12.20.1-buster                                                                                                     0.7s
------
 > [linux/arm64v8 internal] load metadata for docker.io/library/node:12.20.1-buster:
------
error: failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to create LLB definition: no match for platform in manifest sha256:5407853b6aef463dc2da182751015db96958ebe5ea6ccd3bd35fd3f77f03bf8a: not found
```

Using `$TARGETARCH` resolved this issue as well.

apache.Dockerfile
akurinnoy: build fails for `linux/arm64v8` if I don't specify `$TARGETARCH`.

apache.Dockerfile
benoitf: and using `docker.io:12.20.1-alpine3.12` ?

akurinnoy: let me try

akurinnoy:
@benoitf 

without $TARGETARCH:
https://github.com/eclipse/che-dashboard/runs/1813724326?check_suite_focus=true

olexii4 (approved):

benoitf (approved):

sleshchenko (approved):

PR #152, opened 2021-02-02
Render warning message as html

JPinkney:

### What does this PR do?
This PR makes it so that anything in the warning message is rendered as html

~~CQ: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23018~~ ended up not being needed

https://github.com/eclipse/che/issues/18769

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
     "redux-thunk": "^2.3.0",
     "reflect-metadata": "^0.1.13",
     "reselect": "^4.0.0",
+    "sanitize-html": "^2.3.2",
     "sockjs-client": "^1.4.0",
     "umd-compat-loader": "2.1.1",
     "yaml-language-server": "^0.13.0"
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "redux-thunk": "^2.3.0",
     "reflect-metadata": "^0.1.13",
     "reselect": "^4.0.0",
-    "sanitize-html": "^2.3.2",
+    "sanitize-html": "2.3.2",
     "sockjs-client": "^1.4.0",
     "umd-compat-loader": "2.1.1",
     "yaml-language-server": "^0.13.0"
--- a/src/components/BannerAlert/WarningBanner/index.tsx
+++ b/src/components/BannerAlert/WarningBanner/index.tsx
@@ -29,7 +29,7 @@ class WarningBannerAlert extends React.PureComponent<Props, State> {
 
     return (
       <Banner className="pf-u-text-align-center" variant="warning">
-        {warningMessage}
+        <div dangerouslySetInnerHTML={{ __html: warningMessage }}></div>
       </Banner>
     );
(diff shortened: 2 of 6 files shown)


akurinnoy (approved):

olexii4 (approved):

src/components/BannerAlert/WarningBanner/index.tsx
```diff
@@ -29,7 +29,7 @@ class WarningBannerAlert extends React.PureComponent<Props, State> {
 
     return (
       <Banner className="pf-u-text-align-center" variant="warning">
-        {warningMessage}
+        <div dangerouslySetInnerHTML={{ __html: warningMessage }}></div>
```
sleshchenko, line 32: I'm OK with that as simplest option to go, assuming that the cluster admin is responsible for it's content. Then this is not more dangerous than have an ability to modify css, scripts on the packaging, or directly when you do kubectl exec )

sleshchenko (approved):

src/components/BannerAlert/WarningBanner/index.tsx
sleshchenko: Well, later we will introduce an easy option to modify it for Che, CRW on CustomResource instance, but again, everybody who is able edit CheCluster CR - could do anything with custom dockerimage.

akurinnoy: We could sanitize the message from all tags except those we want to permit, using library like this one: https://www.npmjs.com/package/sanitize-html-react

sleshchenko:
Maybe it makes sense to describe such an ability in https://github.com/eclipse/che-dashboard#configurability

ibuziuk (approved):
LGTM, indeed having a PR for docs would be great before merging this PR

JPinkney:
Just waiting for the CQ: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23018 then we can merge. `<a>` tags are allowed and `target` and `href` attributes are allowed, is there other tags/attributes needed?

Ohrimenko1988:
[dashboard-next-ci-test]

sleshchenko:
After https://github.com/eclipse/che-dashboard/pull/154 CQ is not needed anymore >
```diff
+| [`sanitize-html@2.3.2`](https://github.com/apostrophecms/sanitize-html.git) | MIT | clearlydefined |
```
Please regenerate deps files. after you pulled my changes.

package.json
```diff
@@ -71,6 +71,7 @@
     "redux-thunk": "^2.3.0",
     "reflect-metadata": "^0.1.13",
     "reselect": "^4.0.0",
+    "sanitize-html": "^2.3.2",
```
sleshchenko, line 74: I see that we have a lot of ^ but it seems to be a mistake. We must update our dependencies manually, after making sure that's OK with CQs.

PR #155, opened 2021-02-04
Update readme with header.warning explanation

JPinkney:

### What does this PR do?
Adds in header.warning explanation to the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,18 @@ Example:
 }
 ```
 
+Field `"header.warning"` allows you to display a message at the top of the dashboard. You can use HTML to configure the field but only a tags and href, target properties are accepted. It's undefined by default.
+
+Example:
+
+```json
+{
+  "header": {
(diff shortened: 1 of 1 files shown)


sleshchenko (approved):

olexii4 (approved):

README.md
```diff
@@ -79,6 +79,18 @@ Example:
 }
 ```
 
+Field `"header.warning"` allows you to display a message at the top of the dashboard. You can use HTML to configure the field but only a tags and href, target properties are accepted. It's undefined by default.
```
akurinnoy, line 82: I think it will be easier to read if highlight the tag and properties as inline code:

```suggestion
Field `"header.warning"` allows you to display a message at the top of the dashboard. You can use HTML to configure the field but only `<a>` tag and `href`, `target` properties are accepted. It's undefined by default.
```

akurinnoy (approved):

PR #159, opened 2021-02-10
Wrap dashboard pages with error boundary

akurinnoy:
### What does this PR do?

Wrapping UD page components with error boundary to handle uncaught exceptions that may happen in runtime in components. 

**Screenshots**

Screenshot 2021-02-10 at 14 56 54

*expanded component stack*
Screenshot 2021-02-10 at 14 57 46

fixes https://github.com/eclipse/che/issues/18758

### DIFF
--- a/src/services/keycloak/setup.ts
+++ b/src/services/keycloak/setup.ts
@@ -75,16 +75,6 @@ export class KeycloakSetupService {
       const result = await this.testApiAttainability<che.User>(endpoint);
       this.user = result;
     } catch (e) {
-      if (this.isDevelopment) {
-        console.error('Failed to get response to API endpoint: \n' + e);
-        this.user = {
-          email: 'johndoe@test.com',
-          id: 'john-doe-id',
-          name: 'John Doe',
-          links: []
-        };
-        return;
-      }
       throw new Error('Failed to get response to API endpoint: \n' + e);
     }
   }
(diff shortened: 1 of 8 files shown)


src/services/keycloak/setup.ts
```diff
@@ -142,10 +132,6 @@ export class KeycloakSetupService {
       if (e.response.status === 404) {
         return;
       }
-      if (this.isDevelopment) {
```
olexii4, line 145: It won't be caught with ErrorBoundary.  I suggest leaving as is.
These changes do not apply to the target commit.

olexii4 (approved):
LGTM
Please, fix the feedback.

JPinkney (approved):

sleshchenko:
When devfile registry returns invalid JSON as index.json file, the browser logs: `Failed to fetch devfiles metadata from registry URL: https://devfile-registry-che.192.168.39.241.nip.io/devfiles/index.json,TypeError: (intermediate value).data.map is not a function` but user sees it as no devfile is present in the devfile registry.
It would be cool to display error message for a user, like the configured devfile registry is invalid.

^ it's the bug that we face previously. Now dashboard behaves better, since previously blank page was loaded.

src/services/keycloak/setup.ts
akurinnoy: The exception gets handled [here](https://github.com/eclipse/che-dashboard/blob/cc2e9fbd672993d9ef853cc2114b549aab4accc9/src/index.tsx#L35) and then the dashboard continues to initialize. I tested it running dashboard in dev mode either locally and in Che. I may miss something though, if you think so please elaborate on this.

sleshchenko (approved):
The described flow works as expected. I did not find any other corner cases yet but in any case it should be good to go, and we'll handle other errors if any separately

Screenshot_20210211_102804

PR #162, opened 2021-02-11
Cover FactoryLoader with tests

olexii4:

### What does this PR do?
Cover FactoryLoader with tests

fixed https://github.com/eclipse/che/issues/19011


package.json
```diff
@@ -36,7 +36,7 @@
     "licenseCheck:run": "license-tool/run.sh"
   },
   "dependencies": {
-    "@eclipse-che/workspace-client": "0.0.1-1613117389",
+    "@eclipse-che/workspace-client": "^0.0.1-1613117389",
```
sleshchenko, line 39: Is it safe to use ^ here?
It means that newer incompatible version can be fetched, right?

src/store/__mocks__/storeBuilder.ts
```diff
@@ -164,7 +164,7 @@ export class FakeStoreBuilder {
       this.state.workspaces.workspaces = Object.assign([], options.workspaces);
     }
     if (options.workspacesLogs) {
-      this.state.workspaces.workspacesLogs = Object.assign({}, options.workspacesLogs);
+      this.state.workspaces.workspacesLogs = options.workspacesLogs;
```
akurinnoy, line 167: Shallow copy of a map may be done like this:

```suggestion
      this.state.workspaces.workspacesLogs = new Map(options.workspaceLogs);
```

olexii4: Ok

akurinnoy (approved):

package.json
sleshchenko: We agreed that ^ is better to have to let yarn resolve transitive dependencies in a right way.
Autoupgrade should not happen after version is fixed in yarn.lock.
The only thing that should be improved from our-end - we should introduce PR check that guarantees that no dependencies upgrade happens without the corresponding CQ check

olexii4: ^version -  "Compatible with version", will update you to all future minor/patch versions, without incrementing the major version.
But it works before creating a yarn.lock file and will not be upgrading versions each time.

PR #165, opened 2021-02-12
Use the latest workspace client with logs publishing fix

sleshchenko:
### What does this PR do?
This PR upgrades to the latest workspace client with logs publishing fix.

Fixes https://github.com/eclipse/che/issues/19028

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "licenseCheck:run": "license-tool/run.sh"
   },
   "dependencies": {
-    "@eclipse-che/workspace-client": "^0.0.1-1612278649",
+    "@eclipse-che/workspace-client": "0.0.1-1613117389",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -36,7 +36,7 @@
     "licenseCheck:run": "license-tool/run.sh"
   },
   "dependencies": {
-    "@eclipse-che/workspace-client": "^0.0.1-1612278649",
```
sleshchenko, line 39: note that I remove ^ since we need to avoid auto upgrade of dependencies, let me know if I'm wrong

akurinnoy (approved):

olexii4 (approved):

sleshchenko:
Tested it works fine, even verbose mode works good =)

PR #166, opened 2021-02-15
updated github workflows

akurinnoy:
Signed-off-by: Oleksii Kurinnyi <okurinny@redhat.com>

### What does this PR do?

The CI workflow that utilized the Docker Build & Push Action to build multiarch image is rather unstable that prevents to propagate latest changes into next tag. Hense, the CI workflow is updated in order to build image for amd64 architecture only. 

Same time the Docker Build & Push Action used in PR workflow to build simultaneously single arch images for four platforms works well. This pull request adds new workflow intended to build a multiarch image using same approach. 

https://github.com/eclipse/che/issues/19049

### DIFF
--- a/.github/workflows/ci-multiarch.yaml
+++ b/.github/workflows/ci-multiarch.yaml
@@ -1,0 +1,116 @@
+#
+# Copyright (c) 2020 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+name: CI:multiarch
+
+on:
+  workflow_dispatch:
+    inputs: {}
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-multiarch.yaml
```diff
@@ -0,0 +1,116 @@
...
+# SPDX-License-Identifier: EPL-2.0
+#
+
+name: CI:multiarch
+
+on:
+  workflow_dispatch:
+    inputs: {}
+  # push:
+  pull_request:
```
sleshchenko, line 16: it's going to be push again before merging, right?

sleshchenko (approved):

.github/workflows/ci-multiarch.yaml
akurinnoy: correct

skabashnyuk (approved):

olexii4 (approved):

PR #168, opened 2021-02-16
Fix incorrect number of workspaces in confirmation window

akurinnoy:
### What does this PR do?

This PR fixes bug that causes to show in confirmation window an incorrect number of workspaces to be deleted.

fixes https://github.com/eclipse/che/issues/19057


JPinkney (approved):
Just tested and everything seems to be working as expected

sleshchenko (approved):
Works as expected

PR #170, opened 2021-02-17
Add factory policies support

olexii4:

### What does this PR do?
Add factory policies support: `perclick`, `peruser`.
Note `peraccount` is not migrated since it does not really make sense after deprecating organizations.

This PR implements a similar algorithm to the previous version of the dashboard.

For example: ```http://che-eclipse-che.192.168.49.2.nip.io/f?url=https://github.com/eclipse/che-dashboard.git&policies.create=peruser```

fixed https://github.com/eclipse/che/issues/18961

### DIFF
--- a/src/containers/FactoryLoader.tsx
+++ b/src/containers/FactoryLoader.tsx
@@ -261,7 +261,6 @@ export class FactoryLoaderContainer extends React.PureComponent<Props, State> {
       this.showAlert('Failed to create a workspace.');
       return undefined;
     }
-    this.props.setWorkspaceId(workspace.id);
     // check if it ephemeral
     if (workspace.devfile.attributes &&
       workspace.devfile.attributes.persistVolumes === 'false' &&
@@ -274,20 +273,23 @@ export class FactoryLoaderContainer extends React.PureComponent<Props, State> {
     return workspace;
   }
 
-  private async startWorkspace(workspace: che.Workspace): Promise<void> {
+  private async startWorkspace(): Promise<void> {
+    const workspace = this.props.workspace;
+    if (!workspace) {
+      return;
+    }
     if (this.state.currentStep !== LoadFactorySteps.START_WORKSPACE
       && this.state.currentStep !== LoadFactorySteps.OPEN_IDE) {
-      this.setState({ currentStep: LoadFactorySteps.START_WORKSPACE });
-      if (WorkspaceStatus[workspace.status] === WorkspaceStatus.RUNNING) {
-        return;
-      } else {
-        try {
-          await this.props.startWorkspace(`${workspace.id}`);
-        } catch (e) {
-          const workspaceName = workspace.devfile.metadata.name;
-          this.showAlert(`Workspace ${workspaceName} failed to start. ${e.message ? e.message : ''}`);
-          return;
+      try {
+        await this.props.requestWorkspace(workspace.id);
+        if (WorkspaceStatus[workspace.status] === WorkspaceStatus.STOPPED) {
+          await this.props.startWorkspace(workspace.id);
+          this.setState({ currentStep: LoadFactorySteps.START_WORKSPACE });
+        } else if (WorkspaceStatus[workspace.status] === WorkspaceStatus.RUNNING) {
+          this.setState({ currentStep: LoadFactorySteps.START_WORKSPACE });
         }
+      } catch (e) {
+        this.showAlert(`Getting workspace detail data failed. ${e}`);
       }
     }
   }
@@ -332,7 +334,10 @@ export class FactoryLoaderContainer extends React.PureComponent<Props, State> {
     if (!workspace) {
       return;
     }
-    await this.startWorkspace(workspace);
+
+    this.props.setWorkspaceId(workspace.id);
+
+    await this.startWorkspace();
 
     await this.openIde();
   }
--- a/src/containers/FactoryLoader.tsx
+++ b/src/containers/FactoryLoader.tsx
@@ -44,7 +44,7 @@ export enum LoadFactorySteps {
 }
 
 type State = {
-  search: string;
+  search?: string;
   location?: string;
   devfileLocationInfo?: string;
   currentStep: LoadFactorySteps;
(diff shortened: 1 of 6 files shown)


src/containers/FactoryLoader.tsx
```diff
@@ -24,7 +24,9 @@ import { WorkspaceStatus } from '../services/helpers/types';
 import { sanitizeLocation } from '../services/helpers/location';
 import { merge } from 'lodash';
 
-const WS_ATTRIBUTES_TO_SAVE: string[] = ['workspaceDeploymentLabels', 'workspaceDeploymentAnnotations'];
+const WS_ATTRIBUTES_TO_SAVE: string[] = ['workspaceDeploymentLabels', 'workspaceDeploymentAnnotations', 'policies.create'];
+
+type CreatePolicy = 'perclick' | 'peruser' | 'peraccount';
```
sleshchenko, line 29: I have a difficulty with understanding peruser and peraccount...

Per account - means per user or per organization depending on the workspace owner.

1. When I do 
`http://localhost:3000/#/load-factory?url=https://github.com/eclipse/che-dashboard.git&policies.create=peruser` it creates a workspace and reuse it after I open it over and over.

2. After I do `http://localhost:3000/#/load-factory?url=https://github.com/eclipse/che-dashboard.git&policies.create=peraccount`, My workspace is still created under my personal account, the new one, the one from step 1 is not reused.

As far as I see that's not documented and since we deprecated organizations for time being, maybe it would be better to go with perclick, peruser at this point.

Maybe I'm just missing something. I'll read the code tomorrow to understand what peraccount actually does.

src/containers/FactoryLoader.tsx
```diff
@@ -56,10 +59,12 @@ export class FactoryLoaderContainer extends React.PureComponent<Props, State> {
     super(props);
 
     const { search } = this.props.history.location;
+    const createPolicy = 'perclick';
```
sleshchenko, line 62: is it defaultCreatePolicy
```suggestion
    const defaultCreatePolicy = 'perclick';
```
?

src/containers/FactoryLoader.tsx
```diff
@@ -184,12 +197,13 @@ export class FactoryLoaderContainer extends React.PureComponent<Props, State> {
       this.updateOverrideParams(key, val);
       factoryUrl.searchParams.append(key, val);
     });
-
     attrs.stackName = factoryUrl.toString();
+    const createPolicy = attrs['policies.create'] as CreatePolicy || 'perclick';
```
sleshchenko, line 201: I would move 'perclick' to a constant with default value, not sure if it's the way to go in typescript ;)

src/containers/FactoryLoader.tsx
```diff
@@ -215,13 +229,42 @@ export class FactoryLoaderContainer extends React.PureComponent<Props, State> {
...
+      });
+      if (!workspace) {
+        workspace = this.props.allWorkspaces.find(workspace => {
+          return workspace.attributes && workspace.attributes.stackName === attrs.stackName;
+        });
+      }
+    } else if (this.state.createPolicy === 'peraccount') {
+      workspace = this.props.allWorkspaces.find(workspace => {
+        return WorkspaceStatus[workspace.status] === WorkspaceStatus.RUNNING &&
+          devfile && devfile.metadata && devfile.metadata.name === workspace.devfile.metadata.name;
```
sleshchenko, line 246: I don't think it's the right peraccount policy.

The only difference between perUser and perAccount,
perUser should loop over the current user workspaces
perAccount should loop over chosen account, could be useful for organizations, but at this point the only available account - is personal, so should not be any difference with preuser.

BTW I think we should be fine to remove peraccount for timebeing.

src/containers/FactoryLoader.tsx
```diff
@@ -143,6 +141,21 @@ export class FactoryLoaderContainer extends React.PureComponent<Props, State> {
...
+      return;
+    }
+    this.setState({ currentStep: LoadFactorySteps.OPEN_IDE });
+    try {
+      await this.props.requestWorkspace(workspace.id);
+    } catch (e) {
+      this.showAlert(`Getting workspace detail data failed. ${e}`);
+    }
+    await delay();
+    history.push(`/ide/${workspace.namespace}/${workspace.devfile.metadata.name}`);
```
akurinnoy, line 156: We have the helper functions that provide the correct paths for many routes in the application as well as for IDE to open a workspace:`buildIdeLoaderPath`

```suggestion
    history.push(buildIdeLoaderPath(workspace));
```

src/containers/FactoryLoader.tsx
akurinnoy, line 201: These attributes are inputs we should validate. I guess you can create and use a type guard function for this purpose.

akurinnoy: If I understood correctly, it's not necessary because the `createPolicy` constant is treated as of `CreatePolicy` type and allows only defined values to be set.

src/containers/FactoryLoader.tsx
olexii4: I have fixed it.

src/containers/FactoryLoader.tsx
```diff
@@ -215,13 +230,32 @@ export class FactoryLoaderContainer extends React.PureComponent<Props, State> {
     this.setState({ currentStep: LoadFactorySteps.APPLYING_DEVFILE });
     await delay();
 
-    let workspace: che.Workspace | null = null;
-    try {
-      workspace = await this.props.createWorkspaceFromDevfile(devfile, undefined, undefined, attrs) as any;
-    } catch (e) {
-      this.showAlert(`Failed to create a workspace. ${e}`);
+    let workspace: che.Workspace | undefined | null = null;
```
akurinnoy, line 233: Why `null`?

```suggestion
    let workspace: che.Workspace | undefined;
```

olexii4: Ok. I have fixed it.

src/containers/FactoryLoader.tsx
olexii4: Ok. I have fixed it.

src/containers/FactoryLoader.tsx
olexii4: Ok. I have fixed it.

src/containers/FactoryLoader.tsx
olexii4: I have removed 'peraccount' at all.

src/containers/FactoryLoader.tsx
olexii4: Ok. I have fixed it.

src/containers/FactoryLoader.tsx
olexii4: I have removed 'peraccount' at all.

src/containers/FactoryLoader.tsx
```diff
@@ -143,6 +142,25 @@ export class FactoryLoaderContainer extends React.PureComponent<Props, State> {
...
+      await this.props.requestWorkspace(workspace.id);
+    } catch (e) {
+      this.showAlert(`Getting workspace detail data failed. ${e}`);
+    }
+    await delay();
+    history.push(buildIdeLoaderPath(workspace));
+  }
+
+  private getCreatePolicy(attrs: { [key: string]: string }): CreatePolicy {
+    return attrs['policies.create'] as CreatePolicy || DEFAULT_CREATE_POLICY;
```
akurinnoy, line 161: Here you assume that `attrs['policies.create']` has a value that you expect. But it's not necessarily true, there may be any string value. I suggested you add a type guard function similar to described in [typescript handbook](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards)

src/containers/FactoryLoader.tsx
```diff
@@ -215,13 +234,32 @@ export class FactoryLoaderContainer extends React.PureComponent<Props, State> {
...
+    }
+    if (workspace && WorkspaceStatus[workspace.status] === WorkspaceStatus.RUNNING) {
+      this.props.setWorkspaceId(workspace.id);
+      this.setState({ currentStep: LoadFactorySteps.START_WORKSPACE });
+      await this.openIde();
       return;
     }
+    if (!workspace) {
+      try {
+        workspace = await this.props.createWorkspaceFromDevfile(devfile, undefined, undefined, attrs) as any;
```
akurinnoy, line 257: I seem casting result to type `any` is not necessary:

```suggestion
        workspace = await this.props.createWorkspaceFromDevfile(devfile, undefined, undefined, attrs);
```

src/containers/__tests__/FactoryLoader.spec.tsx
```diff
@@ -88,21 +92,25 @@ jest.mock('../../pages/FactoryLoader', () => {
 });
 
 describe('Factory Loader container', () => {
-  const location = 'http://test-location/';
+  const location = 'http://test-location';
   const workspace = createFakeWorkspaceWithRuntime('id-wksp-test');
+  if (workspace.attributes) {
+    workspace.attributes.stackName = 'http://test-location/?policies.create=peruser';
+  }
```
akurinnoy, line 99: How do you know the attribute would be set in a test where you supposed to have it? I think is it's better to do a single-test-specific setup where it's needed, within that particular test.

olexii4: I prepared a dummy store with one default workspace which is used in the one test. Could you explain the benefits of changing it?

dmytro-ndp:
[dashboard-next-ci-test]

src/containers/__tests__/FactoryLoader.spec.tsx
olexii4: I have fixed it.

src/containers/FactoryLoader.tsx
olexii4: I have done it.
Screenshot from 2021-02-22 16-49-47

src/containers/FactoryLoader.tsx
```diff
@@ -143,88 +142,124 @@ export class FactoryLoaderContainer extends React.PureComponent<Props, State> {
...
+  }
+
+  private isCreatePolicy(val: string): val is CreatePolicy {
+    return val && (val as CreatePolicy) === 'perclick' || (val as CreatePolicy) === 'peruser';
+  }
+
+  private getCreatePolicy(attrs: { [key: string]: string }): CreatePolicy | undefined {
+    const policy = attrs['policies.create'] || DEFAULT_CREATE_POLICY;
+    if (this.isCreatePolicy(policy)) {
+      return policy as CreatePolicy;
```
akurinnoy, line 167:
```suggestion
      return policy;
```

olexii4: I have fixed it.

sleshchenko (approved):
Tested and everything works as expected,

apart from one bug that peruser workspace is not opened correctly when there is existing STARTING one

Screenshot_20210223_111930

src/containers/FactoryLoader.tsx
```diff
@@ -143,88 +142,124 @@ export class FactoryLoaderContainer extends React.PureComponent<Props, State> {
...
     }
+    return search;
+  }
 
-    this.setState({ currentStep: LoadFactorySteps.CREATE_WORKSPACE });
-
-    const searchParam = new window.URLSearchParams(location.search);
-    this.resetOverrideParams();
-    const factoryLink = searchParam.get('url');
+  private getLocation(search: string): string | null {
```
akurinnoy, line 192:
```suggestion
  private getLocation(search: string): string | undefined {
```

akurinnoy (approved):

sleshchenko:
And in addition let me leave kudo for the made refactoring :+1:

src/containers/FactoryLoader.tsx
olexii4: I have fixed it

olexii4:
@sleshchenko I have fixed it

PR #171, opened 2021-02-18
fix preload data queue

olexii4:

### What does this PR do?
This PR fixed preload data queue.

fixed https://github.com/eclipse/che/issues/19020

### DIFF
--- a/src/services/bootstrap/PreloadData.ts
+++ b/src/services/bootstrap/PreloadData.ts
@@ -52,21 +52,25 @@ export class PreloadData {
     this.defineEnvironment();
 
     await this.updateUser();
-    await this.updateUserProfile();
-    await this.updateBranding();
-    new ResourceFetcherService().prefetchResources(this.store.getState());
-
-    this.updateRestApiClient();
-    this.updateJsonRpcMasterApi();
-
-    this.updateWorkspaces();
-    this.updateInfrastructureNamespaces();
+    await this.updateJsonRpcMasterApi();
 
     const settings = await this.updateWorkspaceSettings();
-    await this.updatePlugins(settings);
-    await this.updateRegistriesMetadata(settings);
-    await this.updateDevfileSchema();
-    await this.updateUserPreferences();
+
+    await Promise.all([
+      Promise.all([
+        this.updateBranding(),
+        this.updateWorkspaces(),
+        this.updateInfrastructureNamespaces(),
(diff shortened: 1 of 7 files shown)


src/services/bootstrap/PreloadData.ts
```diff
@@ -52,21 +52,25 @@ export class PreloadData {
...
 
     await this.updateUser();
-    await this.updateUserProfile();
-    await this.updateBranding();
-    new ResourceFetcherService().prefetchResources(this.store.getState());
+    await this.updateJsonRpcMasterApi();
 
-    this.updateRestApiClient();
-    this.updateJsonRpcMasterApi();
+    const settings = await this.updateWorkspaceSettings();
```
akurinnoy, line 57: Not sure I got it why you get rid of `updateRestApiClient()` in favor of lazily initialize it, but haven't done the same for `updateJsonRpcMasterApi()`?

src/services/bootstrap/PreloadData.ts
```diff
@@ -52,21 +52,25 @@ export class PreloadData {
...
+        this.updateWorkspaces(),
+        this.updateInfrastructureNamespaces(),
+        this.updateUserProfile(),
+      ]),
+      Promise.all([
+        this.updatePlugins(settings),
+        this.updateRegistriesMetadata(settings),
+        this.updateDevfileSchema(),
+      ])
+    ]);
```
akurinnoy, line 71: Have you considered using [Promise.allSettled()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) here? And also I wonder why do you group methods this way?

src/services/bootstrap/PreloadData.ts
olexii4: I haven't use lazy initialization for ```updateJsonRpcMasterApi()``` because in the existing implementation it ignores a Promise.

akurinnoy (approved):

src/services/bootstrap/PreloadData.ts
olexii4: I was thinking about  ```Promise.allSettled()``` here, but I need a promise that rejects after branding data request fails. So, I decide to use ```Promise.all() ``` instead .

dmytro-ndp:
[dashboard-next-ci-test]

PR #172, opened 2021-02-18
Fix ability to refresh token on API calls

akurinnoy:
### What does this PR do?

This PR fixes the issue when API calls use an expired token. It improves the application pre-start set up so the `che-workspace-client` gets initialized in the correct order, after the keycloak adapter setup. That allows `che-workspace-client` to correctly register a request interceptor that does token refreshing.

fixes https://github.com/eclipse/che/issues/19058

### DIFF
--- a/src/services/cheWorkspaceClient/index.ts
+++ b/src/services/cheWorkspaceClient/index.ts
@@ -62,7 +62,6 @@ export class CheWorkspaceClient {
       }
 
       this.axios.interceptors.request.use(async config => {
-        console.log('>>> interceptor');
         await this.handleRefreshToken(VALIDITY_TIME, config);
         return config;
       });
--- a/src/services/cheWorkspaceClient/index.ts
+++ b/src/services/cheWorkspaceClient/index.ts
@@ -11,10 +11,11 @@
  */
 
 import { AxiosInstance, AxiosRequestConfig } from 'axios';
-import { injectable } from 'inversify';
+import { inject, injectable } from 'inversify';
 import WorkspaceClient, { IWorkspaceMasterApi, IRemoteAPI } from '@eclipse-che/workspace-client';
 import { KeycloakAuthService } from '../keycloak/auth';
 import { EventEmitter } from 'events';
+import { KeycloakSetupService } from '../keycloak/setup';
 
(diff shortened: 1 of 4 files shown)


src/services/cheWorkspaceClient/index.ts
```diff
@@ -53,39 +56,30 @@ export class CheWorkspaceClient {
...
 
-    window.addEventListener('message', (event: MessageEvent) => {
-      if (event.data.startsWith('update-token:')) {
-        const receivedValue = parseInt(event.data.split(':')[1], 10);
-        const validityTime = Number.isNaN(receivedValue) ? VALIDITY_TIME : Math.ceil(receivedValue / 1000);
-        this.handleRefreshToken(validityTime);
-      }
-    }, false);
+      this.axios.interceptors.request.use(async config => {
+        console.log('>>> interceptor');
```
sleshchenko, line 65:
```suggestion
```

sleshchenko (commented):
I even did not try to review the code, but let me just left <3 here for solving issue with using expired token.

According to discussion we had, it's going to refresh token before each request, which is not the most optimal way, but I'm OK with that since it's really stable and you told old dashboard did in the same way.
We can optimize it later if needed.

I'll test it a bit later today

akurinnoy:
@sleshchenko I've looked into token refreshing and found out that requests for updating token are sent only when it's necessary and the token is nearly to expire, not each time when an update method is called. So we don't have to optimize this part.

olexii4 (approved):
LGTM

JPinkney (approved):

dmytro-ndp:
[dashboard-next-ci-test]

dmytro-ndp:
[dashboard-next-ci-test]

sleshchenko:
> @sleshchenko I've looked into token refreshing and found out that requests for updating token are sent only when it's necessary and the token is nearly to expire, not each time when an update method is called. So we don't have to optimize this part.

That's great news!!!
Can we update token forcibly before Start A Workpace request?
It could help Che Server to minimize a chance to use expired token during workspace start process. If it's not done yet, I did not tested or checked the code.

dmytro-ndp:
[dashboard-next-ci-test]

PR #173, opened 2021-02-19
Fix opening workspace details/IDE after it's created out of the current tab

olexii4:

This PR provides an additional API request in the case if we cannot find the target workspace in the storage. These additional requests fixed errors with opening workspace details/IDE pages after it's created out of the current tab.


sleshchenko (commented):
Since it's a pretty urgent fix, please don't wait my approval here

@akurinnoy approval should be enough

akurinnoy (approved):
LGTM

JPinkney (approved):

PR #174, opened 2021-02-23
Fix refresh token callback

olexii4:
### What does this PR do?
Fix refresh token callback in single-user mode.

### How to test it?
Deploy Single User Che with the following CheCluster patch
```yaml
# save the content to /tmp/che-custom-che-server.yaml and execute the following:
# chectl server:deploy --installer=operator --platform=openshift --che-operator-cr-patch-yaml=/tmp/che-custom-che-server.yaml
spec:
  server:
    customCheProperties:
      CHE_MULTIUSER: "false"
```

fixes https://github.com/eclipse/che/issues/19038

### DIFF
--- a/src/services/cheWorkspaceClient/index.ts
+++ b/src/services/cheWorkspaceClient/index.ts
@@ -155,29 +155,27 @@ export class CheWorkspaceClient {
     return Array.from(this._failingWebSockets);
   }
 
-  private async refreshToken(minValidity: number, config?: AxiosRequestConfig): Promise<string | Error> {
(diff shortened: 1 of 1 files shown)


sleshchenko (commented):
I know nothing about KeycloakAuthService, I rely on other reviewers approvals.

sleshchenko (approved):
Tested and it fixes issues with single user, I was able to create workspaces with get started page and then start it.

akurinnoy (approved):

PR #177, opened 2021-02-24
set multiarch container as default, disable...

nickboldt:
### What does this PR do?

set multiarch container as default, disable single-arch build

Change-Id: I3bf13419e30b11b520b4243dbb21b0cf6374547f
Signed-off-by: nickboldt <nboldt@redhat.com>

N/A

N/A (or see commit message above for issue number)

### How to test this PR?
N/A

### PR Checklist

[As the author of this Pull Request I made sure that:](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#pull-request-template-and-its-checklist)

- [x] [The Eclipse Contributor Agreement is valid](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#the-eclipse-contributor-agreement-is-valid)
- [x] [Code produced is complete](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#code-produced-is-complete)
- [x] [Code builds without errors](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#code-builds-without-errors)
- [ ] [Tests are covering the bugfix](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#tests-are-covering-the-bugfix)
- [x] [The repository devfile is up to date and works, if one exists](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#the-repository-devfile-is-up-to-date-and-works)
- [x] [Sections What issues does this PR fix or reference and How to test this PR completed](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#sections-what-issues-does-this-pr-fix-or-reference-and-how-to-test-this-pr-completed)
- [x] [Relevant user documentation updated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#relevant-contributing-documentation-updated)
- [x] [Relevant contributing documentation updated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#relevant-contributing-documentation-updated)
- [x] [CI/CD changes implemented, documented and communicated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#cicd-changes-implemented-documented-and-communicated)

### Reviewers

Reviewers, please comment how you tested the PR when approving it.


nickboldt:
Question:

should we move from 

`CACHE_IMAGE_FULL: docker.io/cheincubator/che-dashboard:cache` 

to new repo in existing quay che-incubator org?

`CACHE_IMAGE_FULL: quay.io/che-incubator/che-dashboard:cache`

That way we're not using as part of the build process, when we're ultimately pushing to Quay anyway.

.github/workflows/ci-multiarch.yaml
```diff
@@ -111,5 +111,5 @@ jobs:
             echo "[!] The job 'build-images' didn't provide any outputs. Can't create the manifest list."
             exit 1;
           fi
-          docker manifest create ${{ env.IMAGE }}:multiarch-next $AMEND
-          docker manifest push ${{ env.IMAGE }}:multiarch-next
+          docker manifest create ${{ env.IMAGE }}:next $AMEND
+          docker manifest push ${{ env.IMAGE }}:next
```
sleshchenko, line 115: Is it OK for default next tag to have arch optionally?

Hard to tell possible cause but GitHub action can push, let's say, only arm64 image...
My point is if multiarch is pushed to default, then we must guarantee that every platform is build before push, or at least `amd64` which is dependency for most cases, which is not the case currently.

.github/workflows/ci.yaml
```diff
@@ -12,16 +12,18 @@ name: CI
 on:
   workflow_dispatch:
     inputs: {}
-  push:
-    branches:
-      - master
+  # disabled as we have ci-multiarch.yaml now
```
sleshchenko, line 15: Well, it not a good motivation to disable single-arch since we used to have multiarch build only, but it was not stable.

Now the situation is better and multiarch build is reworked from buildx:platforms which we used to use https://github.com/eclipse/che-dashboard/pull/166/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddL57

But there is a risk that multiarch build(more often s390x) will fail without any guilt from Dashboard repo side.
On Che Community I raised a topic to check if somebody is interested in multiarch build, and no-one complained, so it's why we decided to keep next as the default and CI-Multiarch is optional.

I don't see a need in amd64-next, if we don't need it, then let's remove.

sleshchenko (commented):
Nobody complained about removing multi-arch build, it's why it was made as optional due stability reasons.

I see potential value in multiarch build as default but github workflow should be reworked in the corresponding way where each (or some of) architectures are mandatory.
And then I don't see a value in keeping single arch image if Che Server and deployments depends on multiarch.

So, in general I'm OK with proposed multiarch by default, but if it won't be stable again, we will need to review it.

.github/workflows/ci-multiarch.yaml
nickboldt: yes, we should be able to quietly warn if s390x fails, but loudly error if any other arch is missing.

nickboldt: > Nobody complained about removing multi-arch build

true... but there's no operator CSV which looks for `:multiarch-next` tags so there's no obvious/easy way to even use these multiarch images. If we use :next then at least the nightly CSV would refer to them and we'd get community testing / PR checks / HP tests etc.

sleshchenko: Then let's just leave multiarch

nickboldt: does that mean "keep the :multiarch-next` tag? 

or does that mean "default to `:next` being multiarch so we immediately expose problems as they happen, and actually use multiarch-built-bits for bleeding edge testing?

sleshchenko: I meant

> "default to :next being multiarch so we immediately expose problems as they happen, and actually use multiarch-built-bits for bleeding edge testing?

.github/workflows/ci.yaml
nickboldt: ok, deleted the file .github/workflows/ci.yaml

.github/workflows/ci-multiarch.yaml
```diff
@@ -111,5 +111,5 @@ jobs:
             echo "[!] The job 'build-images' didn't provide any outputs. Can't create the manifest list."
             exit 1;
           fi
-          docker manifest create ${{ env.IMAGE }}:multiarch-next $AMEND
-          docker manifest push ${{ env.IMAGE }}:multiarch-next
+          docker manifest create ${{ env.IMAGE }}:next $AMEND
```
sleshchenko, line 114: it could lead to case where we push manifest with e.g. PPC64LE only.
We need to make all arch required, or all except s390x (s390x sorry).

Maybe something like https://github.com/eclipse/che-dashboard/commit/be438ae369a6797a09e987236b5e5b26f057416f will work but I did not test.

sleshchenko: just removing `continue-on-error: true` may be alternative approach https://github.com/eclipse/che-dashboard/pull/177/files#diff-0cd88fa8de73cf3bb8b3a02015ac5ff103a7b5a4ea442e75bab856fe9a51ea08L27

dmytro-ndp:
[dashboard-next-ci-test]

dmytro-ndp:
[dashboard-next-ci-test]

.github/workflows/ci-multiarch.yaml
sleshchenko: disabled continue-on-error

.github/workflows/next-build-multiarch.yml
```diff
@@ -106,5 +104,5 @@ jobs:
             echo "[!] The job 'build-images' didn't provide any outputs. Can't create the manifest list."
             exit 1;
           fi
-          docker manifest create ${{ env.IMAGE }}:multiarch-next $AMEND
-          docker manifest push ${{ env.IMAGE }}:multiarch-next
+          docker manifest create ${{ env.IMAGE }}:next $AMEND
```
sleshchenko, line 107: Well, it also requires adapting PR check not to pass when multiarch build fails.
And it has a drawback... For some reason amd64 build takes 4-6 minutes, while other 35-45 minutes which is not critical but not pleasure when you add a rewording fix in the PR before merging.

akurinnoy (approved):

PR #180, opened 2021-03-02
increase the max length of the password allowed for a container regis…

dymart:
…try. This allows for a container registry that uses json keys

kind/usability

### What does this PR do?
Increase the max password length for a container registry. Some container registries don't use passwords but use keys that can be much longer than 100. This pr hopes to make it easier to use these type of container registries. 

This was change was proposed before in the old dashboard.

https://github.com/eclipse/che/issues/5678
https://cloud.google.com/container-registry/docs/advanced-authentication
https://github.com/eclipse/che/pull/13569/files

I'm not sure if the logic should be removed like the previous change or if the max length should be made very high. I proposed a high number to make the change small. If others would like it removed that would also be good.

### DIFF
--- a/src/pages/UserPreferences/RegistryPassword/__tests__/index.spec.tsx
+++ b/src/pages/UserPreferences/RegistryPassword/__tests__/index.spec.tsx
@@ -98,8 +98,8 @@ describe('Registry Password Input', () => {
 
       const input = screen.getByTestId('registry-password-input');
 
-      const message = 'The password is too long. The maximum length is 100 characters.';
-      const allowedPassword = 'a'.repeat(100);
+      const message = 'The password is too long. The maximum length is 10000 characters.';
+      const allowedPassword = 'a'.repeat(10000);
       let label: HTMLElement | null;
 
       userEvent.clear(input);
@@ -108,7 +108,7 @@ describe('Registry Password Input', () => {
       label = screen.queryByText(message);
       expect(label).toBeFalsy();
 
-      const disallowedPassword = 'a'.repeat(101);
+      const disallowedPassword = 'a'.repeat(10001);
 
       userEvent.clear(input);
       userEvent.type(input, disallowedPassword);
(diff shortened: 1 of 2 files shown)


akurinnoy (approved):
LGTM

sleshchenko:
@dymart https://accounts.eclipse.org/legal/eca/validation/61127 is failed, please [sign ECA](https://accounts.eclipse.org/user/eca) and add signoffs in the commits

src/pages/UserPreferences/RegistryPassword/__tests__/index.spec.tsx
```diff
@@ -98,8 +98,8 @@ describe('Registry Password Input', () => {
 
       const input = screen.getByTestId('registry-password-input');
 
-      const message = 'The password is too long. The maximum length is 100 characters.';
-      const allowedPassword = 'a'.repeat(100);
+      const message = 'The password is too long. The maximum length is 10000 characters.';
```
sleshchenko, line 101: such long limit suggest there is no a limit.
I would like to see that logic removed if it's relatively easy

@akurinnoy @olexii4 WDYT?

sleshchenko:
P.S. Thanks for your contribution and sorry for breaking it comparing to old dashboard.

dymart:
Hey @sleshchenko no worries, just noticed it after an update.

dymart:
Hi @sleshchenko, I've signed the ECA and added signoffs to the commits. If there is anything else I need to do let me know. This is my first time contributing here so any direction is appreciated. Thanks

src/pages/UserPreferences/RegistryPassword/__tests__/index.spec.tsx
olexii4: The max length check on the client-side should be in any case.
As for me, 512 characters sounds better. But I do not see any problem with 1000.

olexii4 (approved):
LGTM

sleshchenko (approved):

PR #183, opened 2021-03-09
Accepting private repository factories

akurinnoy:
### What does this PR do?

This PR allows the UD to accept an URL to the private bitbucket repository as a factory. UD opens the OAuth login page if necessary, and then repeats the factory acceptance process.

There is one drawback though, private factory acceptance fails if there aren't pre-created infrastructure namespaces. UD warns the user about the problem with the following message:

`Failed to accept the factory URL. The infrastructure namespace is required to be created. Please create a regular workspace to workaround the issue and open factory URL again.`

The issue is created on Che Server side to handle it correctly on Che Server side https://github.com/eclipse/che/issues/19236.

resolves https://github.com/eclipse/che/issues/18976


src/containers/FactoryLoader.tsx
```diff
@@ -223,15 +229,16 @@ export class FactoryLoaderContainer extends React.PureComponent<Props, State> {
     try {
       await this.props.requestFactoryResolver(location);
     } catch (e) {
-      this.showAlert('Failed to resolve a devfile.');
-      return undefined;
+      if (!isOAuthResponse(e)) {
+        this.showAlert('Failed to resolve a devfile.');
```
sleshchenko, line 233: is it possible to extract the error message from `e` and append it to the message?

akurinnoy: sure, I'll add the fixup

sleshchenko (approved):
LGTM after a fix with notification when namespace does not exist.

Just for history, during testing we together found two places which can be improved, but it's out of the current PR scope:
- The default devfile is resolved from private bitbucket repo after access token is revoked https://github.com/eclipse/che/issues/19273
- Improve workflow when user denies Bitbucket access for Che https://github.com/eclipse/che/issues/19274

akurinnoy:
[dashboard-next-ci-test]

dmytro-ndp:
[dashboard-next-ci-test]

dmytro-ndp:
[dashboard-next-ci-test]

src/containers/FactoryLoader.tsx
```diff
@@ -242,6 +250,43 @@ export class FactoryLoaderContainer extends React.PureComponent<Props, State> {
     return this.getTargetDevfile();
   }
 
+  private resolvePrivateDevfile(oauthUrl: string, location: string): void {
+    try {
+      // looking for a pre-created infrastructure namespace
+      const namespaces = this.props.infrastructureNamespaces.namespaces;
+      if (namespaces.length === 1) {
+        if (!namespaces[0].attributes.phase) {
+          this.showAlert('Failed to accept the factory URL. The infrastructure namespace is required to be created. Please create a regular workspace to workaround the issue and open factory URL again.');
```
skabashnyuk, line 259: I would say that is not an issue. That is by design.

`Please create a regular workspace and open factory URL again` -wdyt?

src/store/FactoryResolver.ts
```diff
@@ -11,13 +11,32 @@
...
+    oauth_provider: string;
+    oauth_version: string;
+    oauth_authentication_url: string;
+  },
+  errorCode: number;
+  message: string;
+}
+
+export function isOAuthResponse(response: any): response is OAuthResponse {
+  if (response?.attributes?.oauth_provider && response?.attributes?.oauth_authentication_url) {
```
olexii4, line 34: Could we simplify it?
```return response?.attributes?.oauth_provider && response?.attributes?.oauth_authentication_url```

olexii4 (approved):
LGTM

JPinkney (approved):
Haven't tested it but the code looks good to me

src/store/FactoryResolver.ts
akurinnoy: I'll provide the fix in my next PR

PR #184, opened 2021-03-10
Force update token on before workspace start

akurinnoy:
Signed-off-by: Oleksii Kurinnyi <okurinny@redhat.com>

### What does this PR do?

This PR forces UD to update the token before performing the start workspace request.

fixes https://github.com/eclipse/che/issues/19045

### DIFF
--- a/src/services/keycloak/auth.ts
+++ b/src/services/keycloak/auth.ts
@@ -22,7 +22,7 @@ export type IKeycloakUserInfo = {
   sub: string;
 };
 
-const MAX_VALIDITY_TIME = 3600;
+const MAX_VALIDITY_TIME_SEC = 3600;
 
 /**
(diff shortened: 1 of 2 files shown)


JPinkney (approved):

src/services/keycloak/auth.ts
```diff
@@ -22,6 +22,8 @@ export type IKeycloakUserInfo = {
   sub: string;
 };
 
+const MAX_VALIDITY_TIME = 3600;
```
sleshchenko, line 25: It's a good practice to make it clear time unit with const or var name, like the following
```suggestion
const MAX_VALIDITY_TIME_MIN = 3600;
```
^ I'm not sure if it's mins or not

sleshchenko (approved):
LGTM just one minor suggestion

PR #191, opened 2021-03-19
Fix listening to workspace logs and statuses

sleshchenko:
### What does this PR do?
Fix listening to workspace logs and statuses.

It fixes the issue that workspace status is not updated when it's started.
Screenshot_20210319_153819

The reason why we must not unsubscribe - unsubscribe does not have scope: 
https://github.com/eclipse/che/blob/master/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/notification/RemoteSubscriptionManager.java#L79

We potentially could stop listening to runtime logs when workspace is started, but it will work until user won't get two starting workspaces, which is possible with configuration.

Issue is created to make workspace client clearer https://github.com/eclipse/che/issues/19349

I've just created on https://github.com/eclipse/che/issues/19348

### DIFF
--- a/src/store/Workspaces/index.ts
+++ b/src/store/Workspaces/index.ts
@@ -181,9 +181,6 @@ function onStatusUpdateReceived(
       status,
     });
   }
-  if (!isDevWorkspace(workspace) && WorkspaceStatus[WorkspaceStatus.STARTING] !== status) {
-    unSubscribeToEnvironmentOutput(workspace.id);
-  }
 }
 
 function subscribeToStatusChange(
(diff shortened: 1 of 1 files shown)


akurinnoy (approved):

olexii4 (approved):

sleshchenko:
Merging to fix the described issue but the dedicated issue will be created to improve a few places:
- we need to unsubscribe before tab close to avoid memory leaks on Che Server. It's just best effort and no guarantee for Che Server that we'll unsubscribe cause of preClose events handles nature.
- we need to remove handlers on client side(events, startup logs) for removed workspace, again to avoid memory leak which should not be significant in any case.

sleshchenko:
[dashboard-next-ci-test]

sleshchenko:
[dashboard-next-ci-test]

sleshchenko:
It seems to be happy path instability but not the current PR changes 
```
      1) Wait for workspace readiness
  [ERROR] CheReporter runner.on.fail: Java Maven test Create Java Maven workspace Wait for workspace readiness failed after 184814ms


  2 passing (3m)
  1 failing

  1) Java Maven test
       Create Java Maven workspace
         Wait for workspace readiness:
     TimeoutError: Exceeded the maximum number of checking attempts, project has not been imported
      at ProjectTree.waitProjectImported (/tmp/e2e/pageobjects/ide/ProjectTree.ts:266:15)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

PR #194, opened 2021-03-25
Improve release script

mkuznyetsov:

### What does this PR do?
- Add option to force override tags during release
- Send MM notification when this job fails

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,6 +7,10 @@ on:
         description: 'The version that is going to be released. Should be in format 7.y.z'
         required: true
         default: '7.y.z'
+      forceRecreateTags:
+        description: If true, tags will be recreated. Use with caution
+        required: false
+        default: 'false'
 
 env:
   IMAGE: quay.io/eclipse/che-dashboard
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -104,17 +108,41 @@ jobs:
       -
         name: "Checkout Che Dashboard source code"
         uses: actions/checkout@v2
+
+      -
+        name: Check existing tags
+        run: |
+          set +e
+          RECREATE_TAGS=${{ github.event.inputs.forceRecreateTags }}
```
sleshchenko, line 116: We have pretty much copy pasting. It may be a good time to introduce out github action. How sure how easy/difficult it's to implement.
Maybe we may just encapsulate release script in dockerimage.

.github/workflows/release.yml
```diff
@@ -104,17 +108,41 @@ jobs:
...
-      - name: Create MM message
+      - name: Create failure MM message
+        if: ${{ failure() }}
+        run: |
+          echo "{\"text\":\":building_construction: Che Dashboard ${{ github.event.inputs.version }} release has failed: https://github.com/eclipse/che-dashboard/actions/workflows/release.yml\"}" > mattermost.json
+      - name: Create success MM message
         run: |
           echo "{\"text\":\":building_construction: Che Dashboard ${{ github.event.inputs.version }} has been released: https://quay.io/eclipse/che-dashboard:${{ github.event.inputs.version }}\"}" > mattermost.json
       - name: Send MM message
+        if: ${{ success() }} || ${{ failure() }}
```
sleshchenko, line 145: Can this if statement be false?

sleshchenko (approved):

.github/workflows/release.yml
mkuznyetsov: if workflow is cancelled, it will have different status. So I guessed, that we don't want to send notification in that instance

sleshchenko: makes sense

PR #197, opened 2021-03-26
Fix apiVersion in ownerReference so that devworkspacetemplates cleaned up

JPinkney:

### What does this PR do?
This PR fixes the devworkspacetemplate ownerReference so that it is set correctly

https://github.com/eclipse/che/issues/19328

### DIFF
--- a/src/services/workspace-client/devWorkspaceClient.ts
+++ b/src/services/workspace-client/devWorkspaceClient.ts
@@ -13,7 +13,7 @@
 import { inject, injectable } from 'inversify';
 import { convertDevWorkspaceV2ToV1, isDeleting, isWebTerminal } from '../helpers/devworkspace';
 import { WorkspaceClient } from './';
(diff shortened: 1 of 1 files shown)


sleshchenko (approved):

src/services/workspace-client/devWorkspaceClient.ts
```diff
@@ -93,15 +93,16 @@ export class DevWorkspaceClient extends WorkspaceClient {
       // todo handle error in a proper way
       const pluginName = pluginDevfile.metadata.name.replaceAll(' ', '-').toLowerCase();
       const workspaceId = createdWorkspace.status.workspaceId;
+      const devfileGroupVersion = 'workspace.devfile.io/v1alpha2';
```
sleshchenko, line 96: I just wonder if this constant should be imported from devworkspace client?

PR #198, opened 2021-03-30
Enhance error handling for when che workspace routing isn't installed

JPinkney:

### What does this PR do?
This PR makes it so that a devworkspace will error when a devworkspace is stuck on `'Waiting for workspace routing objects'`. This is because when the devworkspace che controller isn't installed the devworkspace-controller loops and never get's past the starting state. It's not ideal to depend on the exact message from the devworkspace controller but I don't really see any other way.

To test:
```
echo "spec:
  devWorkspace:
    enable: true" > devworkspace-config.yaml
chectl server:deploy --installer=operator --platform=openshift --che-operator-cr-patch-yaml=devworkspace-config.yaml
```
then scale down both deployments in the devworkspace-che namespace and then create a devworkspace with:
```
#{CHE_URL}/#/load-factory/?url=https://github.com/l0rd/spring-petclinic/tree/devfile2
```

https://github.com/eclipse/che/issues/19325

### DIFF
--- a/src/services/workspace-client/devWorkspaceClient.ts
+++ b/src/services/workspace-client/devWorkspaceClient.ts
@@ -80,8 +80,6 @@ export class DevWorkspaceClient extends WorkspaceClient {
       throw new Error(`Could not retrieve devworkspace status information from ${workspaceName} in namespace ${namespace}`);
     } else if (workspaceStatus.phase.toUpperCase() === DevWorkspaceStatus[DevWorkspaceStatus.RUNNING] && !workspaceStatus?.ideUrl) {
       throw new Error('Could not retrieve ideUrl for the running workspace');
-    } else if (workspaceStatus.message === 'Waiting for workspace routing objects') {
-      throw new Error(workspaceStatus.message);
     }
     return convertDevWorkspaceV2ToV1(workspace);
   }
--- a/src/services/workspace-client/devWorkspaceClient.ts
+++ b/src/services/workspace-client/devWorkspaceClient.ts
@@ -75,8 +75,13 @@ export class DevWorkspaceClient extends WorkspaceClient {
       await delay();
     }
     this.checkForDevWorkspaceError(workspace);
-    if (!workspace.status || !workspace.status.phase || !workspace.status.ideUrl) {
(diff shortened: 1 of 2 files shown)


src/services/workspace-client/devWorkspaceClient.ts
```diff
@@ -75,8 +75,13 @@ export class DevWorkspaceClient extends WorkspaceClient {
       await delay();
     }
     this.checkForDevWorkspaceError(workspace);
-    if (!workspace.status || !workspace.status.phase || !workspace.status.ideUrl) {
+    const workspaceStatus = workspace.status;
+    if (!workspaceStatus || !workspaceStatus.phase) {
       throw new Error(`Could not retrieve devworkspace status information from ${workspaceName} in namespace ${namespace}`);
+    } else if (workspaceStatus.phase.toUpperCase() === DevWorkspaceStatus[DevWorkspaceStatus.RUNNING] && !workspaceStatus?.ideUrl) {
+      throw new Error('Could not retrieve ideUrl for the running workspace');
+    } else if (workspaceStatus.message === 'Waiting for workspace routing objects') {
```
sleshchenko, line 83: how much we wait before checking this?

JPinkney: It's currently 10 seconds. We can increase it but then we are increasing the time it takes for the factory flow to finish, even on workspaces that wouldn't run into this issue

sleshchenko: I would vote for removing this check for time being.
Once we propagate the phase to logs, it should not be a big issue.
So, workspace will be in starting phase forever if Che Routing is not installed properly.

JPinkney: Do you mean remove the `waiting for workspace routing objects` check? Should I keep the ideUrl check or do you think I should just close this PR?

sleshchenko (approved):
LGTM after removing error in case `Waiting for workspace routing`.
After we better expose workspace startup, it should be good enough.

sleshchenko (approved):

PR #212, opened 2021-04-14
Move to pnp/yarn 2

olexii4:

### What does this PR do?
Move to pnp/yarn 2

resolves https://github.com/eclipse/che/issues/19613


src/Layout/Navigation/__tests__/MainItem.spec.tsx
```diff
@@ -20,8 +20,12 @@ import { NavigationItemObject } from '..';
 import { FakeStoreBuilder } from '../../../store/__mocks__/storeBuilder';
 import { Provider } from 'react-redux';
 
+const _expect = expect as any;
```
benoitf, line 23: hello, I'm curious why we need to do that ?

olexii4: @benoitf  We need to do that because of wrong types. Somehow it resolved in a wrong way after migration.
I added that as a quick fix.

benoitf: ok thanks

akurinnoy: Hmm, I've just migrated the dashboard to yarn2 from scratch and all tests passed as is. No fixes are necessary.

akurinnoy: it's because I've put `nodeLinker: node-modules` into my `.yarnr.yml`, while `PnP` breaks tests.

.yarn/build-state.yml
```diff
@@ -0,0 +1,18 @@
+# Warning: This file is automatically generated. Removing it is fine, but will
```
merceyz, line 1: This file shouldn't be commited

src/Layout/Navigation/__tests__/MainItem.spec.tsx
merceyz: Hey, just cloned and checked, the type of `expect` is `jest.expect` so it works with PnP

.yarn/build-state.yml
olexii4: Thanks. It was added before I update the '.gitignore' file.
Fixed

src/Layout/Navigation/__tests__/MainItem.spec.tsx
olexii4: @merceyz Thanks a lot

olexii4: Fixed

merceyz:
Since you're using PnP you should look at https://yarnpkg.com/getting-started/migration#editor-support so vscode works

olexii4:
@merceyz Thanks.  I have fixed docker build.

dmytro-ndp:
[dashboard-ci-test]

package.json
```diff
@@ -28,13 +28,12 @@
...
     "test:coverage": "jest --coverage",
-    "test:watch": "jest --watch",
-    "licenseCheck:prepare": "license-tool/build.sh",
-    "licenseCheck:run": "license-tool/run.sh"
+    "test:watch": "jest --watch"
   },
   "dependencies": {
     "@eclipse-che/devworkspace-client": "^0.0.1-1617871695",
     "@eclipse-che/workspace-client": "^0.0.1-1615287725",
+    "@patternfly/patternfly": "~4.84.0",
```
akurinnoy, line 36: This dependency is not necessary so should be removed along with this line:
https://github.com/eclipse-che/che-dashboard/blob/main/src/index.tsx#L18

src/components/DevfileEditor/__tests__/DevfileEditor.spec.tsx
```diff
@@ -17,10 +17,53 @@ import DevfileEditor from '../';
...
+    asDiagnostics = jest.fn();
+  }
+  class MonacoToProtocolConverter {
+    asPosition = jest.fn();
+    asCompletionItem = jest.fn();
+  }
+  return {
+    ProtocolToMonacoConverter,
+    MonacoToProtocolConverter
+  };
```
akurinnoy, line 66: These changes seem redundant to me, the test only has to be slightly updated to pass. See https://github.com/eclipse-che/che-dashboard/blob/yarn2-migration/src/components/DevfileEditor/__tests__/DevfileEditor.spec.tsx#L22-L24

src/components/DevfileEditor/index.tsx
```diff
@@ -14,8 +14,8 @@ import React from 'react';
 import { connect, ConnectedProps } from 'react-redux';
 import { AppState } from '../../store';
 import { DisposableCollection } from '../../services/helpers/disposable';
-import { ProtocolToMonacoConverter, MonacoToProtocolConverter } from 'monaco-languageclient/lib/monaco-converter';
-import { languages, IEditorConstructionOptions } from 'monaco-editor-core/esm/vs/editor/editor.main';
+import { ProtocolToMonacoConverter, MonacoToProtocolConverter } from './monaco-converter';
```
akurinnoy, line 17: I don't see how changes in this component are related to moving the project to yarn2. Could you point out why they are needed?

src/components/DevfileEditor/index.tsx
```diff
@@ -243,8 +244,8 @@ export class DevfileEditor extends React.PureComponent<Props, State> {
     );
   }
 
-  private getDecorations(): monaco.editor.IModelDecoration[] {
-    const decorations: monaco.editor.IModelDecoration[] = [];
+  private getDecorations(): any[] {
+    const decorations: any[] = [];
```
akurinnoy, line 248: Please take this into account: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#any

src/components/DevfileEditor/__tests__/DevfileEditor.spec.tsx
olexii4: Existing implementation 

```
import { languages, editor } from 'monaco-editor-core/esm/vs/editor/editor.main';

jest.mock('../../../../node_modules/monaco-editor-core/esm/vs/editor/editor.main', () => {
  return () => ({ languages, editor });
});
```
was done in a wrong way. I just fixed it.

Do you have another opinion about it?

akurinnoy: It will be really nice if you provide more details about what was wrong.

package.json
olexii4: We need it because of this import 
```
import '@patternfly/patternfly/patternfly-addons.css';
```
https://github.com/eclipse-che/che-dashboard/blob/main/src/index.tsx#L19

We need '@patternfly/patternfly' because of import 'patternfly-addons.css'

Am I right?

src/components/DevfileEditor/__tests__/DevfileEditor.spec.tsx
akurinnoy: maybe, it would be even better to move these changes into a separate PR for a more convenient discussion.

package.json
akurinnoy: yep, I've pointed the wrong line. 

We needed that import because of not proper usage of the patterfly components. Now we're OK with that, so the import of the css file and the dependency should be deleted.

src/components/DevfileEditor/index.tsx
olexii4: Good catch. I forgot to change back. I have fixed it.

src/components/DevfileEditor/index.tsx
olexii4: It depends how yarn 2 resolve packages. Old import was broken. I did not find a good solution(except 'nodeLinker: node-modules' - is not a good idea as for me) for fix it and decide that the best one will be just copy the target file add a link.

@akurinnoy  If you have some better idea, please, provide a PR with changes.

src/components/DevfileEditor/__tests__/DevfileEditor.spec.tsx
olexii4: I just provide existing mock in a right way because the problem has been identified after migrating to yarn 2.
Let discuss it here(this jest.mock for 'monaco-editor-core/esm/vs/editor/editor.main').

olexii4: The problem: 
```
import { languages, editor } from 'monaco-editor-core/esm/vs/editor/editor.main';

jest.mock('../../../../node_modules/monaco-editor-core/esm/vs/editor/editor.main', () => {
  return () => ({ languages, editor });
});
```
existing mock for  tried to get values from themselves.

It was a bug.

akurinnoy: Now I got it, thank you for the explanation.

src/components/DevfileEditor/index.tsx
akurinnoy: When my local build was failing on this file, adding this resolution to the package.json fixed it for me:

```json
  "resolutions": {
    ...
    "vscode-languageserver-protocol": "3.15.3"
  }
```

you may have a different error though.

package.json
olexii4: I have fixed it

src/components/DevfileEditor/index.tsx
olexii4: Thanks all. I have fixed it.

akurinnoy (approved):

sleshchenko (approved):
During testing, I faced the issue that we report `undefined` when a user is not able to start more workspaces, but I think the same happens in main branch as well.

LGTM Good job!

PR #219, opened 2021-04-20
Add Che Server functionality for devfile v1 overrides

olexii4:

### What does this PR do?
Add Che Server functionality for devfile v1 overrides.
Screenshot from 2021-04-19 17-55-07

fixed https://github.com/eclipse/che/issues/19424

### DIFF
--- a/src/containers/FactoryLoader.tsx
+++ b/src/containers/FactoryLoader.tsx
@@ -22,13 +22,11 @@ import FactoryLoader from '../pages/FactoryLoader';
 import { selectAllWorkspaces, selectPreferredStorageType, selectWorkspaceById } from '../store/Workspaces/selectors';
 import { WorkspaceStatus } from '../services/helpers/types';
 import { buildIdeLoaderPath, sanitizeLocation } from '../services/helpers/location';
-import { merge } from 'lodash';
 import { lazyInject } from '../inversify.config';
 import { KeycloakAuthService } from '../services/keycloak/auth';
 import { getEnvironment, isDevEnvironment } from '../services/helpers/environment';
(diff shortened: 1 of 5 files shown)


src/containers/FactoryLoader.tsx
```diff
@@ -84,21 +84,10 @@ export class FactoryLoaderContainer extends React.PureComponent<Props, State> {
...
-        overrideKeys.push(regExpExecArray[0]);
-      }
-      if (overrideKeys.length > 0) {
-        let currentVal = this.overrideDevfileObject;
-        for (let index = 1; index < overrideKeys.length; index++) {
-          currentVal[overrideKeys[index]] = index === overrideKeys.length - 1 ? val : {};
-          currentVal = currentVal[overrideKeys[index]];
-        }
-      }
+    if (!key.startsWith('override.')) {
```
sleshchenko, line 87: Do you think it makes sense to keep this private func after you modified it to be really tiny? or better to inline if into https://github.com/eclipse-che/che-dashboard/blob/CHE-19424/src/containers/FactoryLoader.tsx#L220

I prefer a bit the second option

sleshchenko (approved):
I haven't tested yet but changes LGTM

akurinnoy (approved):

src/containers/FactoryLoader.tsx
olexii4: I have fixed it

olexii4: https://github.com/eclipse-che/che-dashboard/pull/219/files#diff-4a8e21344754b4ee47f503546e5d899120cc756884f5b1c3b266399ddc3ec003R217

PR #222, opened 2021-04-21
Don't hide the navbar if it's not the IDE page

akurinnoy:
### What does this PR do?

This PR prevents unwanted hiding the navbar, which happens sometimes when a user navigates from Theia IDE that is starting to any other UD page. Now the UD checks if it's the IDE page and only then hides the navbar.


sleshchenko (approved):
Changes seems meaningful to me)

olexii4 (approved):

PR #223, opened 2021-04-21
Improve ErrorBoundary component to handle resources unavailability error

akurinnoy:
### What does this PR do?

The ErrorBoundary component will handle when resources are not available and refresh the dashboard after the timeout. Users will be able to prevent page reloading or reload it manually.

https://user-images.githubusercontent.com/16220722/115908569-0232a580-a473-11eb-9c62-9f34a2ec471f.mov

fixes https://github.com/eclipse/che/issues/19504


src/Layout/ErrorBoundary/index.tsx
```diff
@@ -55,50 +65,137 @@ export class ErrorBoundary extends React.PureComponent<Props, State> {
...
+      return;
+    }
+
+    return (
+      <Alert
+        isInline
+        variant={AlertVariant.warning}
+        title={
+          <React.Fragment>
+            <span>This page is reloading in&nbsp;</span>
```
sleshchenko, line 158: I think we need to make it more clearer for the user why we're refreshing a page:
```
The application has been likely updated on the server. Refreshing a page to get newer resources in .
Contact an administrator if refreshing continues after the next load.
```
^ can we use local storage to detect when page is endless (2-3 times) refreshed automatically, to simplify the message for user and show the last part only after we detect it?

src/Layout/ErrorBoundary/index.tsx
```diff
@@ -55,50 +65,137 @@ export class ErrorBoundary extends React.PureComponent<Props, State> {
...
+      this.startCountdown();
+    }
+  }
+
+  public componentWillUnmount(): void {
+    this.toDispose.dispose();
+  }
+
+  private testResourceNotFound(error: Error): boolean {
+    return /loading chunk [\d]+ failed/i.test(error.message);
```
sleshchenko, line 82: I wonder if we can merge this info with `refreshing banner` under View details, not to make users panic in such a pretty regular case.

sleshchenko (commented):
That's much more better than now but something can be still improved, I hope.

src/Layout/ErrorBoundary/index.tsx
akurinnoy: sure, that's a really nice improvement

src/Layout/ErrorBoundary/index.tsx
akurinnoy: I'm not sure the banner fits here because this is the runtime error that breaks the UD, while the banner is used when the UD is functional and the user needs to be notified that something happened.

olexii4 (approved):
LGTM
I have tested it.
Screenshot from 2021-04-22 11-38-03
Maybe it will be better to remove "Stop countdown" action after stopping.

akurinnoy:
updated the screencast in the description

sleshchenko (approved):
screencast looks good to me

JPinkney (approved):

sleshchenko:
[dashboard-next-ci-test]

sleshchenko:
[dashboard-next-ci-test]

sleshchenko:
[dashboard-next-ci-test]

sleshchenko:
[dashboard-next-ci-test]

sleshchenko:
[dashboard-next-ci-test]

PR #224, opened 2021-04-21
Fix token refresher for single-user

olexii4:
### What does this PR do?
Fix token refresher for single-user.

fixes https://github.com/eclipse/che/issues/19499

### DIFF
--- a/src/Layout/index.tsx
+++ b/src/Layout/index.tsx
@@ -108,12 +108,11 @@ export class Layout extends React.PureComponent<Props, State> {
           isHeaderVisible: true,
         });
       } else if (event.data === 'hide-navbar') {
-        if (this.testIdePath()) {
-          this.setState({
-            isSidebarVisible: false,
-            isHeaderVisible: false,
-          });
-        }
+        const isHeaderVisible = !this.testIdePath() || document.getElementById('ide-iframe') === null;
+        this.setState({
+          isSidebarVisible: false,
+          isHeaderVisible,
+        });
       }
     };
     window.addEventListener('message', handleMessage, false);
--- a/src/Layout/index.tsx
+++ b/src/Layout/index.tsx
@@ -102,19 +102,18 @@ export class Layout extends React.PureComponent<Props, State> {
         return;
       }
 
-      if (this.testIdePath() === false) {
-        return;
-      }
       if (event.data === 'show-navbar') {
         this.setState({
           isSidebarVisible: true,
           isHeaderVisible: true,
         });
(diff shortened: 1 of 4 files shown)


sleshchenko (approved):

src/Layout/index.tsx
```diff
@@ -102,19 +102,18 @@ export class Layout extends React.PureComponent<Props, State> {
         return;
       }
 
-      if (this.testIdePath() === false) {
```
akurinnoy, line 105: Could you please elaborate on why you decided not to prevent showing the navbar?

My thoughts about this are the IDE iframe should not control the navbar visibility until the IDE loader page is opened. Because that may be the user's decision to hide the navbar at the moment, and this change will cause the issue similar to [described here](https://github.com/eclipse-che/che-dashboard/pull/222#issue-620036582) with the only difference that the navbar will be unexpectedly shown.

akurinnoy (changes requested):

src/Layout/index.tsx
olexii4: I decided to do it because of two thought: 1. Running script can have a delay and we should live an ability to show navbar. 2. An issue which bring that changes was about hiding navbar only. 

Could you reproduce an issue when navbar will be unexpectedly shown? 

With my solution you can always hide navbar with a button, and it is not the same level of problem we can have without 'show-navbar' event.

olexii4: BTW:   These changes added a regression with editor Expanded mode and I tried to minimize it.

For example:

```
    this.handleExpand = () => {
      if (this.state.isExpanded) {
        window.postMessage('show-navbar', '*');
        const isExpanded = false;
        this.setState({ isExpanded });
        this.props.handleExpand(isExpanded);
      } else {
      ...
```

https://github.com/eclipse-che/che-dashboard/blob/main/src/components/LogsTab/LogsTools/index.tsx#L41

olexii4: I have fixed regression
https://github.com/eclipse-che/che-dashboard/pull/224/commits/b31dec274575794f5a9ad689ba4e859715d5cbcc

akurinnoy: Thank you for the explanation, that all definitely make sense.

But this:

> I disagree with the preventing 'show-navbar' event.

makes me a bit frustrated. I suppose you should not have approved [this PR](https://github.com/eclipse-che/che-dashboard/pull/222) if you have any objections to how it's done or how it works.

akurinnoy (approved):

PR #225, opened 2021-04-23
Show finalizing devworkspaces until it's actually gone

olexii4:
Signed-off-by: Oleksii Orel <oorel@redhat.com>

### What does this PR do?
Show finalizing devworkspaces until it's actually gone

### What issues does this PR fix or reference?
fixed https://github.com/eclipse/che/issues/19324

PR #226, opened 2021-04-23
Add in the ability to edit devworkspaces in the dashboard

JPinkney:
### What does this PR do?
This PR makes so it so that you are able to edit devworkspaces in the dashboard and save them to your cluster.
The basic flow with editing devworkspaces is:
- If you are saving a devworkspace and the devworkspace is not running then save it directly to the cluster without asking
- If you are saving a devworkspace and the devworkspace is running then ask for confirmation if they want to restart or not
    - If they say they want to restart then update the devworkspace directly on the cluster and then restart the devworkspace
        - This will also remove the `che.eclipse.org/dashboard-next-run` annotation if it is present and proceed with whatever was directly in the devfile editor 
    - If they say they do not want to restart then add in a `che.eclipse.org/dashboard-next-run` annotation which will store the devworkspace to run next time the devworkspace is started
        - The next time the devworkspace is started if `che.eclipse.org/dashboard-next-run` exists it will patch the devfile of the current workspace with the devfile from `che.eclipse.org/dashboard-next-run` and start the workspace
        - If the `che.eclipse.org/dashboard-next-run` annotation does not exist when starting a devworkspace then just proceed as normal 

Also:
- `che.eclipse.org/dashboard-next-run` is present when opening a devworkspace in the editor, then that is the devfile that will be shown to the user in the editor
- If someone changes the devworkspace on the cluster when you are editing the devfile in the dashboard an error will be thrown and the user editing the devfile in the dashboard will have to reload

restart-workspace

This PR depends on https://github.com/che-incubator/devworkspace-client/pull/21

Fixes https://github.com/eclipse/che/issues/19330

### DIFF
--- a/src/pages/WorkspaceDetails/index.tsx
+++ b/src/pages/WorkspaceDetails/index.tsx
@@ -40,7 +40,7 @@ import { selectIsLoading, selectWorkspaceById } from '../../store/Workspaces/sel
 import { History } from 'history';
 
 import './WorkspaceDetails.styl';
-import { Workspace } from '../../services/workspaceAdapter';
+import { isWorkspaceV1, Workspace } from '../../services/workspaceAdapter';
 
 export const SECTION_THEME = PageSectionVariants.light;
 
@@ -208,7 +208,10 @@ export class WorkspaceDetails extends React.PureComponent<Props, State> {
               <EditorTab
                 ref={this.editorTabPageRef}
                 workspace={workspace}
-                onSave={workspace => this.onSave(workspace)} />
+                onSave={workspace => this.onSave(workspace)}
+                hasDevWorkspaceWarning={() => this.setState({
+                  hasWarningMessage: true
+                })} />
             </Tab>
             {/* <Tab eventKey={WorkspaceDetailsTabs.Logs} title={WorkspaceDetailsTabs[WorkspaceDetailsTabs.Logs]}>*/}
             {/*  <LogsTab workspaceId={workspace.id} />*/}
@@ -238,7 +241,7 @@ export class WorkspaceDetails extends React.PureComponent<Props, State> {
   }
 
   private async onSave(workspace: Workspace): Promise<void> {
-    if (this.props.workspace && (WorkspaceStatus[this.props.workspace.status] !== WorkspaceStatus.STOPPED)) {
+    if (this.props.workspace && (WorkspaceStatus[this.props.workspace.status] !== WorkspaceStatus.STOPPED) && isWorkspaceV1((this.props.workspace as Workspace).ref)) {
       this.setState({ hasWarningMessage: true });
     }
     await this.props.onSave(workspace, this.state.activeTabKey);
--- a/src/pages/WorkspaceDetails/index.tsx
+++ b/src/pages/WorkspaceDetails/index.tsx
@@ -209,7 +209,7 @@ export class WorkspaceDetails extends React.PureComponent<Props, State> {
                 ref={this.editorTabPageRef}
                 workspace={workspace}
                 onSave={workspace => this.onSave(workspace)}
-                hasDevWorkspaceWarning={() => this.setState({
+                onDevWorkspaceWarning={() => this.setState({
                   hasWarningMessage: true
                 })} />
             </Tab>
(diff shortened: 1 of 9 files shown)


src/services/workspace-client/devWorkspaceClient.ts
```diff
@@ -31,6 +31,8 @@ export interface IStatusUpdate {
   workspaceId: string;
 }
 
+export const DEVWORKSPACE_NEXT_START_ANNOTATION = 'che.eclipse.org/dashboard-next-run';
```
sleshchenko, line 34:
```suggestion
/**
 * DevWorkspace Configuration that is postponed till next start to avoid devworkspace restart.
 */
export const DEVWORKSPACE_NEXT_START_ANNOTATION = 'che.eclipse.org/next-start-cfg';
```
^ I assume it should be respected by any client, like chectl.
If we want to make it component specific, it would be `dashboard.che.eclipse.org/...` I think. Like openshift console has `console.openshift.io`

sleshchenko: I wonder if dashboard should flush such delayed configuration to already stopped workspaces? or it would just make stuff more complicated?

sleshchenko (approved):
The described flow works fine for me, apart from invalid handling of failed devworkspaces. But I'm ok with moving it out of the current PR scope, like item n.3 in https://github.com/eclipse/che/issues/19684

@olexii4 changes review is on your end

src/services/workspace-client/devWorkspaceClient.ts
```diff
@@ -141,8 +135,81 @@ export class DevWorkspaceClient extends WorkspaceClient {
...
+    // Take the devworkspace with no plugins and then inject them
+    for (const plugin of plugins) {
+      const pluginName = this.normalizePluginName(plugin.metadata.name, workspace.status.devworkspaceId);
+      this.addPlugin(workspace, pluginName, workspace.metadata.namespace);
+    }
+
+    const namespace = workspace.metadata.namespace;
+    const name = workspace.metadata.name;
+
+    const patch = [] as Patch[];
```
olexii4, line 159: Maybe ```const patch: Patch[] = [];``` 
WDYT?

src/pages/WorkspaceDetails/index.tsx
```diff
@@ -208,7 +208,10 @@ export class WorkspaceDetails extends React.PureComponent<Props, State> {
               <EditorTab
                 ref={this.editorTabPageRef}
                 workspace={workspace}
-                onSave={workspace => this.onSave(workspace)} />
+                onSave={workspace => this.onSave(workspace)}
+                hasDevWorkspaceWarning={() => this.setState({
```
olexii4, line 212: ```hasDevWorkspaceWarning``` -> ```onDevWorkspaceWarning```

```
  <EditorTab
        ref={this.editorTabPageRef}
        workspace={workspace}
        onSave={workspace => this.onSave(workspace)} 
        onDevWorkspaceWarning={() => this.handleOnDevWorkspaceWarning()} />

...

  private handleOnDevWorkspaceWarning() {
      this.setState({
         hasWarningMessage: true
      })
  }

```
WDYT?

olexii4:
Found a bug https://youtu.be/GUxfoYiJdWk

We will have an ability to fix it after https://github.com/eclipse-che/che-dashboard/pull/225

olexii4:
LGTM
The described flow works fine for me. 

If changes review is on my end: @JPinkney you can merge it after fixing small feedback
https://github.com/eclipse-che/che-dashboard/pull/226#discussion_r623028697

olexii4 (approved):

src/services/workspace-client/devWorkspaceClient.ts
JPinkney: Do you mean when the workspace is stopping then if there is something in the configuration apply it directly to the devworkspace then delete the annotation? It might make it a bit more complicated.

When you edit a stopped devworkspace in the dashboard and then click save it will automatically update the devworkspace directly instead of adding the annotation

sleshchenko: I meant flush annotation value to config as soon as possible, when Dashboard detect stopped workspace with the annotation.

It may help to avoid config lose when users edit config with kubectl, when it has dashboard annotation. Then Dashboard will override their changes with next run.
But it's minor, let's discuss and track it separately.

PR #209, opened 2021-04-07
Improve dev workspaces support

akurinnoy:
### What does this PR do?

This PR provides the following changes to improve support of dev workspaces:
- introduced separate redux stores for Che workspaces and dev workspaces, and the Workspaces store now only dispatches actions between them;
- introduced the WorkspaceAdapter class, for React components not to depend on the shape of provided workspace objects.

### What issues does this PR fix or reference?

resolves https://github.com/eclipse/che/issues/19184

PR #227, opened 2021-04-28
Backport adapting dashboard to optional default DW plugins

sleshchenko:
### What does this PR do?
Backports adapting dashboard to optional default DW plugins.
I'm not sure if it prevents workspace from starting, but at least it helps to avoid error state where error is logged into browser console. Probably it works without it, but I haven't checked since I found another issue where CRW 2.8 plugin registry does not return Theia devfile.yaml

Backport https://github.com/eclipse-che/che-dashboard/pull/209/commits/7acb70c0dc5d40d02850fa94f851b3ef491a443a


sleshchenko:
Closing this since CRW works fine without it

PR #250, opened 2021-05-24
Workspace details page improvements

olexii4:

### What does this PR do?
Workspace details page improvements:
 - select a tab depends on the search param;
 - create a new component which will block a route in the case with unsaved changes;

fixed https://issues.redhat.com/browse/CRW-1717


akurinnoy (approved):

src/pages/WorkspaceDetails/index.tsx
```diff
@@ -77,92 +73,88 @@ export class WorkspaceDetails extends React.PureComponent<Props, State> {
...
+      return this.editorTabPageRef.current?.state.hasChanges || !this.editorTabPageRef.current?.state.isDevfileValid;
+    } else if (this.state.activeTabKey === WorkspaceDetailsTab.OVERVIEW) {
+      return this.overviewTabPageRef.current?.hasChanges === true;
+    }
+    return false;
   }
 
   public render(): React.ReactElement {
-    const { workspace, workspacesLink } = this.props;
+    const { workspace, workspacesLink, history } = this.props;
```
sleshchenko, line 157: It works just fine for the described issue but seems I found one more case when it still does not work, it does not work in case of incorrect YAML, just add any symbol after `:`, or remove space between any key: value pair:
Screenshot_20210524_152939

Then you're able to switch to overview without ask to discard changes.
When you return back to devfile editor - your invalid yaml is there, if you haven't applied changes on Overview page.

src/components/UnsavedChangesModal/__tests__/UnsavedChangesModal.spec.tsx
```diff
@@ -0,0 +1,58 @@
...
+    jest.clearAllMocks();
+    jest.useFakeTimers();
+  });
+
+  afterEach(() => {
+    jest.useRealTimers();
+  });
+
+  it('should inform users about unsaved changes', () => {
+    const message = 'You have unsaved changes. You may go ahead and discard all changes, or close this window and save them.';
```
sleshchenko, line 41: @olexii4 Does this PR also fix https://github.com/eclipse/che/issues/19584 ?

olexii4: Yes. It will fix it.

src/pages/WorkspaceDetails/index.tsx
sleshchenko: I've created a separate issue to handle that since it's not bug introduced of this PR https://github.com/eclipse/che/issues/19873

sleshchenko (approved):
:+1:

PR #253, opened 2021-05-26
feat: add support for all available devfile schema versions

JPinkney:

### What does this PR do?
This PR updates the editor in the dashboard to use schema version 2.1.0.
It depends on https://github.com/eclipse-che/che-server/pull/25

### DIFF
--- a/src/store/DevfileRegistries/index.ts
+++ b/src/store/DevfileRegistries/index.ts
@@ -171,11 +171,15 @@ export const actionCreators: ActionCreators = {
         const patchedJSONString = JSON.stringify(schemav1).replaceAll('#/definitions', '#/oneOf/0/definitions');
         const parsedSchemaV1 = JSON.parse(patchedJSONString);
 
-        const schemav2 = await WorkspaceClient.restApiClient.getDevfileSchema('2.0.0');
+        const schemav200 = await WorkspaceClient.restApiClient.getDevfileSchema('2.0.0');
+        const schemav210 = await WorkspaceClient.restApiClient.getDevfileSchema('2.1.0');
+        const schemav220alpha = await WorkspaceClient.restApiClient.getDevfileSchema('2.2.0-alpha');
         schema = {
           oneOf: [
             parsedSchemaV1,
-            schemav2
+            schemav200,
+            schemav210,
+            schemav220alpha
           ]
         };
       }


src/store/DevfileRegistries/index.ts
```diff
@@ -171,7 +171,7 @@ export const actionCreators: ActionCreators = {
         const patchedJSONString = JSON.stringify(schemav1).replaceAll('#/definitions', '#/oneOf/0/definitions');
         const parsedSchemaV1 = JSON.parse(patchedJSONString);
 
-        const schemav2 = await WorkspaceClient.restApiClient.getDevfileSchema('2.0.0');
+        const schemav2 = await WorkspaceClient.restApiClient.getDevfileSchema('2.1.0');
```
sleshchenko, line 174: Well, I think we should support all available schemas, like 2.0.0, 2.1.0, 2.2.0-alpha... WDYT?

sleshchenko: maybe we should request the version according to devfile content...

src/store/DevfileRegistries/index.ts
```diff
@@ -171,11 +171,15 @@ export const actionCreators: ActionCreators = {
         const patchedJSONString = JSON.stringify(schemav1).replaceAll('#/definitions', '#/oneOf/0/definitions');
         const parsedSchemaV1 = JSON.parse(patchedJSONString);
 
-        const schemav2 = await WorkspaceClient.restApiClient.getDevfileSchema('2.0.0');
+        const schemav200 = await WorkspaceClient.restApiClient.getDevfileSchema('2.0.0');
+        const schemav210 = await WorkspaceClient.restApiClient.getDevfileSchema('2.1.0');
```
sleshchenko, line 175: I wonder if devfile schema should set an explicit value for schema version, like here https://github.com/devfile/api/blob/main/schemas/latest/devfile.json#L1560
It would guarantee that devfile if validate with a proper version of the schema.
I'll raise it on the Devfile Cabal

akurinnoy (approved):

src/store/DevfileRegistries/index.ts
```diff
@@ -171,11 +171,15 @@ export const actionCreators: ActionCreators = {
         const patchedJSONString = JSON.stringify(schemav1).replaceAll('#/definitions', '#/oneOf/0/definitions');
         const parsedSchemaV1 = JSON.parse(patchedJSONString);
```
olexii4, line 172: ```parsedSchemaV1``` => ```schemav100```

To be consistent with  
```
            ...
            schemav200,
            schemav210,
            schemav220alpha
            ...
```

olexii4: @JPinkney WDYT?

sleshchenko: I left as is. v1 seems even clearer than v100

PR #261, opened 2021-06-03
Displays devfile v2 getting started if devWorkspace toggle is enabled

benoitf:
### What does this PR do?
Display in getting started the `v2` links of devfile registries if DevWorkspace support is enabled.
image

https://github.com/eclipse/che/issues/19923

### DIFF
--- a/src/store/DevfileRegistries/selectors.ts
+++ b/src/store/DevfileRegistries/selectors.ts
@@ -14,14 +14,21 @@ import { createSelector } from 'reselect';
 import { AppState } from '../';
 import match from '../../services/helpers/filter';
 
-const selectState = (state: AppState) => state.devfileRegistries;
+const selectState = (state: AppState) => ({ devfileRegistries: state.devfileRegistries, workspacesSettings: state.workspacesSettings });
 
 export const selectRegistriesMetadata = createSelector(
   selectState,
   state => {
-    const registriesMetadata = Object.values(state.registries)
+    const registriesMetadata = Object.values(state.devfileRegistries.registries)
       .map(registryMetadata => registryMetadata.metadata || []);
-    return mergeRegistriesMetadata(registriesMetadata);
+    const metadata = mergeRegistriesMetadata(registriesMetadata);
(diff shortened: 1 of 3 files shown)


benoitf:
@olexii4 @akurinnoy if you could tell me if it's using a good approach, what I need to change, follow, etc. (For sure I know that it may require tests but if the global idea is totally wrong then :D )

dmytro-ndp:
[dashboard-ci-test]

src/store/DevfileRegistries/selectors.ts
```diff
@@ -14,22 +14,29 @@ import { createSelector } from 'reselect';
 import { AppState } from '../';
 import match from '../../services/helpers/filter';
 
-const selectState = (state: AppState) => state.devfileRegistries;
+const selectState = (state: AppState) => ({ devfileRegistries: state.devfileRegistries, workspacesSettings: state.workspacesSettings });
```
akurinnoy, line 17: I think of importing the `selectWorkspacesSettingsState` or just creating this state so you don't need to modify other selectors but only `selectRegistriesMetadata`:

```typescript

import { selectWorkspacesSettingsState } from '../Workspaces/Settings/selectors';
// or
const selectWorkspacesSettingsState = (state: AppState) => state.workspacesSettings;

const selectState = (state: AppState) => state.devfileRegistries;

export const selectRegistriesMetadata = createSelector(
  selectState,
  selectWorkspacesSettingsState,
  (devfileRegistriesState, workspacesSettingsState) => {
    const registriesMetadata = Object.values(devfileRegistriesState.registries)
      .map(registryMetadata => registryMetadata.metadata || []);
    const metadata = mergeRegistriesMetadata(registriesMetadata);
    const cheDevworkspaceEnabled = workspacesSettingsState.settings['che.devworkspaces.enabled'] === 'true';
    if (cheDevworkspaceEnabled) {
      return filterDevfileV2Metadata(metadata);
    } else {
      return metadata;
    }

  }
);
```

benoitf: yes it looks far better

benoitf:
closing in favor of https://github.com/eclipse-che/che-dashboard/pull/262 as I should push to repository instead of a fork to have all PR checks green

PR #264, opened 2021-06-07
Add a project if imported when using devfile v2 factory flow

olexii4:

### What does this PR do?
This PR brings several improveents:
- prevent undefined in error handling on devfile resolving;
- provision project if devfile v2 miss it;
- add workaround case with empty devfile v2 volume component;
- add test coverage for getDevfile and getProgectName methods;

fixes https://github.com/eclipse/che/issues/19371


src/typings/che.d.ts
```diff
@@ -115,12 +115,12 @@ declare namespace che {
     attributes: { [propName: string]: string };
   }
 
-  export interface ProjectSource {
-    location: string;
-    parameters?: {
-      [paramName: string]: any;
-    };
-    type?: string;
+  export interface ProjectV2Source {
```
sleshchenko, line 118: Isn't it project source used for Che Workspaces? I would like to see type used from DevWorkspace client or devfile API https://github.com/che-incubator/devworkspace-client/blob/main/src/types.ts#L80

sleshchenko:
[dashboard-ci-test]

src/typings/che.d.ts
olexii4: I have fixed it.

src/services/helpers/types.ts
```diff
@@ -22,11 +22,26 @@ export interface AlertItem {
...
+  'scm_info'?: FactoryResolverScmInfo;
 }
 
+export type FactoryResolverScmInfo = {
+  'clone_url': string;
+  'scm_provider': string;
+  'branch'?: string;
+};
+
+export type ProjectV2Source = {
```
sleshchenko, line 37:
```suggestion
export type DevfileV2ProjectSource = {
```
^ may be a bit clearer but feel free to ignore it, anyway we would need to replace it with type from devfile/api

src/store/helpers.ts
```diff
@@ -18,3 +23,50 @@
...
+    const scmInfo = data['scm_info'];
+    if (scmInfo) {
+      const origin = scmInfo['clone_url'];
+      const name = getProjectName(origin);
+      const revision = scmInfo.branch;
+      const project: ProjectV2Source = { name, git: { remotes: { origin } } };
+      if (revision) {
+        project.git.checkoutFrom = { revision };
+      }
+      projects.push(project);
```
sleshchenko, line 54: well, Do I miss check or no if statement exists that we provision project only in case none is specified in namespace?

src/typings/che.d.ts
```diff
@@ -115,14 +115,6 @@ declare namespace che {
     attributes: { [propName: string]: string };
   }
 
-  export interface ProjectSource {
```
sleshchenko, line 118: Why you remove it, is it only unused?
Please remove all unused here, or none =)

olexii4: It was outdate. Let us do it step by step.

src/store/helpers.ts
olexii4: Yes, if statement exist it will be overridden: 
```devfile = Object.assign({ projects }, devfile);```

File: src/services/helpers/types.ts (hunk 2 above)
Comment (olexii4) [reply src/services/helpers/types.ts]: I have fixed it.

**Comment by olexii4** [comment]
[dashboard-ci-test]

File: src/store/helpers.ts (hunk 3 above)
Comment (olexii4) [reply src/store/helpers.ts:54]: And I have improved it
```
...
if (!devfile.projects?.length && scmInfo) {
...
```

akurinnoy (approved):

sleshchenko (approved):
As far as I tested it worked just fine for me.

PR #265, opened 2021-06-07
feat: Propagate the plugin registry and dashboard urls to the containers in the initial devworkspace templates

JPinkney:
…ers in the initial devworkspace templates

### What does this PR do?
This PR propagates the plugin registry and dashboard urls to every container in the initial devworkspacetemplate

https://github.com/eclipse/che/issues/19933

### DIFF
--- a/src/services/workspace-client/devWorkspaceClient.ts
+++ b/src/services/workspace-client/devWorkspaceClient.ts
@@ -48,6 +48,8 @@ export class DevWorkspaceClient extends WorkspaceClient {
   private initializing: Promise<void>;
   private lastDevWorkspaceLog: Map<string, string>;
   private devWorkspacesIds: string[];
+  private pluginRegistryUrlName: string;
+  private dashboardUrlName: string;
 
   constructor(@inject(KeycloakSetupService) keycloakSetupService: KeycloakSetupService) {
     super(keycloakSetupService);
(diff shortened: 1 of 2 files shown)


benoitf (commented):
What will happen for theia in DWO without Che ? We'll have to give some env for the plugin registry url ?

sleshchenko:
> What will happen for theia in DWO without Che ? We'll have to give some env for the plugin registry url ?

We're not supposed to run Theia in DWO without Che.
If you mean our samples, probably we could manually update some of samples for testing particular Theia functionality, but if we want Theia be able to run without Che, dashboard and plugin registry URLs must be optional.

src/services/workspace-client/devWorkspaceClient.ts
```diff
@@ -48,6 +48,8 @@ export class DevWorkspaceClient extends WorkspaceClient {
   private initializing: Promise<void>;
   private lastDevWorkspaceLog: Map<string, string>;
   private devWorkspacesIds: string[];
+  private pluginRegistryUrlName: string;
+  private dashboardUrlName: string;
```
sleshchenko, line 52:
```suggestion
  private dashboardUrlEnvName: string;
```
and the same above

sleshchenko (approved):
I haven't checked yet but LGTM in general. Just a few minor suggestions

akurinnoy (approved):

olexii4 (approved):

PR #268, opened 2021-06-10
Dashboard backend and frontend integration

akurinnoy:
### What does this PR do?

- moves dashboard backend(is not used yet) and frontend in one monorepo;
- adds static-server to use instead of Apache to serve frontend;
- adds Dockerfile (with dashboard frontend only that is served by a new static server);


packages/dashboard-backend/src/init.ts
```diff
@@ -0,0 +1,58 @@
...
+ * Copyright (c) 2021 Red Hat, Inc.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ ***********************************************************************/
+
+import { INodeConfig } from "@eclipse-che/devworkspace-client";
```
olexii4, line 11: ``` "@eclipse-che/devworkspace-client"``` -> ``` '@eclipse-che/devworkspace-client'````

akurinnoy: This should be another issue to set up linters and formatters for the whole project to fix such issues.

akurinnoy:
@olexii4 @sleshchenko @JPinkney I've force-pushed this branch, please review it again

make-release.sh
```diff
@@ -57,6 +57,16 @@ bump_version () {
   git checkout "${CURRENT_BRANCH}"
 }
 
+function update_pkgs_versions() {
+  lerna version --no-git-tag-version -y "${VERSION}"
```
sleshchenko, line 61: TODO version  should be pass as argument since one place rely on VERSION, one on NEXT_VERSION

packages/dashboard-backend/LICENSE
```diff
@@ -0,0 +1,277 @@
+Eclipse Public License - v 2.0
```
sleshchenko, line 1: Do we need LICENSE in each of package?

packages/dashboard-backend/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "version": "7.33.0-SNAPSHOT",
+  "description": "Dashboard backend for Eclipse Che",
+  "main": "lib/server.js",
+  "types": "lib/index.d.ts",
+  "scripts": {
+    "build": "rimraf lib && webpack-cli --mode development --config webpack.config.js",
+    "prepare": "yarn run build",
+    "test": "jest --passWithNoTests"
+  },
+  "contributors": [
```
sleshchenko, line 12: Don't be shy to add here yourself as well.

packages/dashboard-backend/src/config.ts
```diff
@@ -0,0 +1,15 @@
...
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ ***********************************************************************/
+
+/**
+ * Common configuration variables
+ */
+
+export const routingClass = "che";
```
sleshchenko, line 15:
```suggestion
export const routingClass = "che";

```

packages/dashboard-backend/src/openshift/kubeconfig.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    // Create new kubeconfig and authenticate as the user
+    const kc = new KubeConfig();
+    kc.loadFromClusterAndUser(createCluster(), createUser(openShiftToken));
+    
+    // Use the newly authenticated user
+    nodeApi.config = kc;
+    return nodeApi;
+}
+
```
sleshchenko, line 84:
```suggestion
```

packages/dashboard-backend/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+        "baseUrl": "src",
+        "skipLibCheck": true,
+        "sourceMap": true
+    },
+    "exclude": [
+        "node_modules",
+        "test"
+    ],
+    "include": ["src"]
+}
```
sleshchenko, line 18:
```suggestion
}

```

packages/dashboard-backend/LICENSE
akurinnoy: I suppose we don't need it, but let's do the cleanup after this PR is merged.

sleshchenko (commented):
I looked though changes and they LGTM, just a few minor comments.
I'll test it soon

packages/dashboard-backend/package.json
akurinnoy: I don't think it makes a difference until we don't publish it. I'll consider it though.

olexii4 (approved):
LGTM

JPinkney (approved):
LGTM

sleshchenko (approved):
There is at least one small regression but that's fine to proceed with them since main branch already has 7.34.0-SNAPSHOT so, we have 3 weeks to get it fixed.

PR #271, opened 2021-06-14
feat: get rid of -alpha suffix in schemaVersion for unreleased version

sleshchenko:
### What does this PR do?
Adapts dashboard to getting rid of alpha suffix in schemaVersion for unreleased devfiles.
Depends on https://github.com/eclipse-che/che-server/pull/27

### DIFF
--- a/src/store/DevfileRegistries/index.ts
+++ b/src/store/DevfileRegistries/index.ts
@@ -173,7 +173,7 @@ export const actionCreators: ActionCreators = {
 
         const schemav200 = await WorkspaceClient.restApiClient.getDevfileSchema('2.0.0');
         const schemav210 = await WorkspaceClient.restApiClient.getDevfileSchema('2.1.0');
-        const schemav220alpha = await WorkspaceClient.restApiClient.getDevfileSchema('2.2.0-alpha');
+        const schemav220alpha = await WorkspaceClient.restApiClient.getDevfileSchema('2.2.0');
         schema = {
           oneOf: [
             parsedSchemaV1,


src/store/DevfileRegistries/index.ts
```diff
@@ -173,7 +173,7 @@ export const actionCreators: ActionCreators = {
 
         const schemav200 = await WorkspaceClient.restApiClient.getDevfileSchema('2.0.0');
         const schemav210 = await WorkspaceClient.restApiClient.getDevfileSchema('2.1.0');
-        const schemav220alpha = await WorkspaceClient.restApiClient.getDevfileSchema('2.2.0-alpha');
+        const schemav220alpha = await WorkspaceClient.restApiClient.getDevfileSchema('2.2.0');
```
akurinnoy, line 176: Why shouldn't you rename the variable as well?

amisevsk (approved):

src/store/DevfileRegistries/index.ts
sleshchenko: to make it clearer that it's not released yet and the latest stable to 210

akurinnoy (approved):

benoitf (approved):

src/store/DevfileRegistries/index.ts
benoitf, line 176: maybe rename the constant :-)

sleshchenko: I can do it but then I need a comment that would explain that it's not released yet as I have in Che Server https://github.com/eclipse-che/che-server/pull/27/files#diff-b1166dc0647d4d5da93d2d6753d2281d94182ab9f5a4a91bc58854d1436abdefR31 )

JPinkney (approved):

PR #272, opened 2021-06-14
Try to use types from devfile/api

sleshchenko:
### What does this PR do?
It's failed try to use types from devfile/api https://github.com/che-incubator/devworkspace-client/pull/26
It failed due to pain of optional fields forever even required like metadata.name and namespace.
Also, devworkspace metadata is object.
So, DevWorkspace Client or Dashboard would need their own wrapper to make usage simpler.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "webpack-cli": "^3.3.11"
   },
   "resolutions": {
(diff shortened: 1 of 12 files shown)


sleshchenko:
closing in favor of https://github.com/eclipse-che/che-dashboard/pull/329

PR #282, opened 2021-06-23
feat: set up openshift ci image with chectl installed

sleshchenko:
### What does this PR do?
feat: set up openshift ci image with chectl installed #282

It's for https://github.com/eclipse/che/issues/19982

### DIFF
--- a/.ci/openshift-ci/Dockerfile
+++ b/.ci/openshift-ci/Dockerfile
@@ -1,0 +1,22 @@
+#!/bin/bash
+#
+# Copyright (c) 2012-2020 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+# Dockerfile to bootstrap build and test in openshift-ci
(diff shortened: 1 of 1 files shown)


.ci/openshift-ci/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+# Dockerfile to bootstrap build and test in openshift-ci
+FROM registry.ci.openshift.org/openshift/release:golang-1.15
+SHELL ["/bin/bash", "-c"]
+# Install yq, kubectl, chectl cli used by olm/olm.sh script.
```
sleshchenko, line 15:
```suggestion
# Install yq, kubectl, chectl cli used by olm/olm.sh script.
```
what is olm.sh? Is it a way we're going to install Che for Dashboard on OpenShift CI?

akurinnoy (approved):
LGTM

.ci/openshift-ci/Dockerfile
musienko-maxim: I am thinking.... So , actually it  was grabbed from other project. And I do not know which approach will be more preferable for us for CHE instalation. Actually, I am going to disscuss it with you. It just like init commit for laucnching and debugging the job. It will be changed according to our needs

PR #295, opened 2021-07-12
fix: make dashboard rest k8s like

sleshchenko:
### What does this PR do?
fix: make dashboard rest k8s like. Example:
```
https://$CHE_HOST/api/unsupported/k8s/apis/workspace.devfile.io/v1alpha2/namespaces/opentlc-mgr-che/devworkspaces/quarkus-quickstart
```

N/A

### DIFF
--- a/packages/dashboard-backend/src/index.ts
+++ b/packages/dashboard-backend/src/index.ts
@@ -66,7 +66,7 @@ server.addContentTypeParser(
 );
 
 server.get(
-  "/workspace/namespace/:namespace",
+  "/namespace/:namespace/devworkspaces",
   {
     schema: {
       headers: authenticationHeaderSchema,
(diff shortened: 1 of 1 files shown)


akurinnoy (approved):

olexii4 (approved):

JPinkney (approved):

packages/dashboard-backend/src/index.ts
```diff
@@ -85,7 +85,7 @@ server.get(
 );
 
 server.post(
-  "/workspace",
+  "/namespace/:namespace/devworkspaces",
```
JPinkney, line 88: If you want to use namespace like that you'll need to do something like: https://github.com/eclipse-che/che-dashboard/blob/main/packages/dashboard-backend/src/index.ts#L78 (unless we just ignore namespace here)

sleshchenko: thanks for point it. I assumed I can draft paths and let someone else fix the logic haha.
I'll fix it before merging.

akurinnoy (approved):

PR #296, opened 2021-07-12
Fix changing workspace storage type from the overview tab

akurinnoy:
### What does this PR do?

The PR fixes a regression when changing workspace storage type on the workspace overview tab doesn't change the workspace devfile.

fixes https://github.com/eclipse/che/issues/19939

### DIFF
--- a/packages/dashboard-frontend/src/services/workspaceAdapter/index.ts
+++ b/packages/dashboard-frontend/src/services/workspaceAdapter/index.ts
@@ -200,7 +200,14 @@ export class WorkspaceAdapter<T extends che.Workspace | IDevWorkspace> implement
   set storageType(type: che.WorkspaceStorageType) {
     if (isWorkspaceV1(this.workspace)) {
       const attributes = typeToAttributes(type);
-      Object.assign(this.workspace.devfile.attributes, attributes);
+      if (!this.workspace.devfile.attributes) {
+        this.workspace.devfile.attributes = {};
+      }
+      if (attributes) {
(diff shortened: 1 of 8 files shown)


packages/dashboard-frontend/src/services/workspaceAdapter/index.ts
```diff
@@ -200,7 +200,14 @@ export class WorkspaceAdapter<T extends che.Workspace | IDevWorkspace> implement
   set storageType(type: che.WorkspaceStorageType) {
     if (isWorkspaceV1(this.workspace)) {
       const attributes = typeToAttributes(type);
-      Object.assign(this.workspace.devfile.attributes, attributes);
+      if (!this.workspace.devfile.attributes) {
+        this.workspace.devfile.attributes = {};
+      }
+      if (attributes) {
+        Object.assign(this.workspace.devfile.attributes, attributes);
```
olexii4, line 207: The result of this 'Object.assign(this.workspace.devfile.attributes, attributes)' could bring a wrong storage type'.
It is better to remove 'persistVolumes' and 'asyncPersist' keys, then set needed.

akurinnoy: That makes sense, I'll provide a fix.

akurinnoy: done

olexii4 (approved):

sleshchenko (approved):
I haven't tested but changes seem to be meaningful

PR #298, opened 2021-07-14
chore: Rename github workflow files

mkuznyetsov:
chore: Rename github workflow files #298

### What does this PR do?
- Rename GitHub workflow files for Next build, to share the same name along with Eclipse Che next build jobs
- rename filename extensions yaml -> yml


nickboldt (approved):
+1 for shorter extensions and consistency!

.github/workflows/next-build.yml
```diff
@@ -7,7 +7,7 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 
-name: CI
+name: Next build
```
sleshchenko, line 10: Next DockerimagebIs it convention? I mean using `build` since it may be understood like only build without storing images. In DWO we have it `Next Dockerimage` which may be a bit cleaner.

nickboldt: Would "Next container image" be better then? Docker is a competitor, and a specific implementation (which isn't podman). :)

mkuznyetsov: so I guess we could go with whatever project owners prefer, as this is simply a description (we hadn't come up with conventions for it).

akurinnoy: +1 for "Next container image"

sleshchenko: > Would "Next container image" be better then? Docker is a competitor, and a specific implementation (which isn't podman). :)

good catch. Next container image works for me

nickboldt: > whatever project owners prefer

Nope, we should pick one convention and use that everywhere. What value is there in being inconsistent when the whole rationale to rename files here is to be consistent across the repos? :D

akurinnoy (approved):

.github/workflows/next-build-multiarch.yml
```diff
@@ -5,9 +5,9 @@
 # which is available at https://www.eclipse.org/legal/epl-2.0/
 #
 # SPDX-License-Identifier: EPL-2.0
-#
+#
```
sleshchenko, line 8:
```suggestion
#
```

sleshchenko (approved):
Just one minor suggestion

PR #303, opened 2021-07-21
Prevent creating a devworkspaces if default editor is not fetched

akurinnoy:
### What does this PR do?

This PR improves the devworkspaces creation flow in the way that the Dashboard prevents creating new workspaces if the default editor devfile hasn't been fetched beforehand. Users are still will be able to start existing devworkspaces though.

fixes https://github.com/eclipse/che/issues/19852


olexii4 (approved):
I have tested it. It works as expected.

Знімок екрана 2021-07-21 о 14 57 26

packages/dashboard-frontend/src/services/bootstrap/PreloadData.ts
```diff
@@ -128,8 +129,16 @@ export class PreloadData {
       this.watchNamespaces(defaultNamespace);
     }
 
-    const { requestDwDevfiles } = DwPlugins.actionCreators;
-    await requestDwDevfiles(`${settings.cheWorkspacePluginRegistryUrl}/plugins/${settings['che.factory.default_editor']}/devfile.yaml`)(this.store.dispatch, this.store.getState, undefined);
+    const { requestDwDefaultEditor } = DwPluginsStore.actionCreators;
+    try {
+      await requestDwDefaultEditor(settings)(this.store.dispatch, this.store.getState, undefined);
+    } catch (e) {
+      const message = `Not able to create workspace due required resources failed to load: ${e}`;
```
JPinkney, line 136: I wonder if it makes sense to change this to be something like:
```suggestion
      const message = `Required sources failed when trying to create the workspace: ${e}`;
```
or
```suggestion
      const message = `Unable to create the workspace. Requesting default editor failed: ${e}`;
```

packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
```diff
@@ -332,18 +333,24 @@ export const actionCreators: ActionCreators = {
   },
     pluginRegistryUrl: string | undefined,
   ): AppThunk<KnownAction, Promise<IDevWorkspace>> => async (dispatch, getState): Promise<IDevWorkspace> => {
+
+    const state = getState();
+
+    if (state.dwPlugins.defaultEditorError) {
+      const message = `Not able to create workspace due required resources failed to load: ${state.dwPlugins.defaultEditorError}`;
```
JPinkney, line 340: Same comment as above,

I wonder if it makes sense to change this to be something like:
```suggestion
      const message = `Required sources failed when trying to create the workspace: ${state.dwPlugins.defaultEditorError}`;
```
or
```suggestion
      const message = `Unable to create the workspace. Requesting default editor failed: ${state.dwPlugins.defaultEditorError}`;
```

JPinkney (approved):
Other then my two suggestions it looks good to me

packages/dashboard-frontend/src/services/bootstrap/PreloadData.ts
akurinnoy: Nice suggestions, thanks!

sleshchenko (approved):

PR #305, opened 2021-07-26
fix: provision Che env var into theia generated resources

sleshchenko:
### What does this PR do?
fix: provision Che env var into theia generated resources.

To test it:
1. deploy Che with devworkspace enabled.
2. Create devworkspace.
3. Check that VSX installer has needed env vars
Screenshot_20210726_162554


benoitf (approved):
LGTM as we do the update after all templates are there

PR #310, opened 2021-07-28
fix: Fixes for release script and workflow

mkuznyetsov:

### What does this PR do?
Fixes for Che Dashboard release script
- fix sed that didn't work on linux
- add npm and lerna installation into GH workflow runner

N / A

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -136,8 +136,6 @@ jobs:
           git config --global user.name "Mykhailo Kuznietsov"
           git config --global user.email "mkuznets@redhat.com"
           export GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }}
-
-          npm install --global lerna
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -129,6 +136,8 @@ jobs:
           git config --global user.name "Mykhailo Kuznietsov"
           git config --global user.email "mkuznets@redhat.com"
           export GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }}
+
+          npm install --global lerna
```
sleshchenko, line 140: why do we need it twice?

mkuznyetsov: we don't 😄 removing

sleshchenko (approved):
I haven't tested but LGTM

akurinnoy (approved):

PR #312, opened 2021-07-29
Merge registries metadata

akurinnoy:
### What does this PR do?

This PR makes registries objects metadata to be merged to prevent the cause when the following piece of metadata overwrites the previous one.

fixes https://issues.redhat.com/browse/CRW-2116

### DIFF
--- a/packages/dashboard-frontend/src/store/DevfileRegistries/index.ts
+++ b/packages/dashboard-frontend/src/store/DevfileRegistries/index.ts
@@ -13,7 +13,7 @@
 import { Action, Reducer } from 'redux';
 import { AppThunk } from '..';
 import { fetchRegistryMetadata, fetchDevfile } from '../../services/registry/devfiles';
-import { createState } from '../helpers';
+import { createObject } from '../helpers';
 import { container } from '../../inversify.config';
 import { CheWorkspaceClient } from '../../services/workspace-client/cheWorkspaceClient';
 import { getErrorMessage } from '../../services/helpers/getErrorMessage';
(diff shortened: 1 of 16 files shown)


olexii4 (approved):

packages/dashboard-frontend/src/store/DevfileRegistries/index.ts
```diff
@@ -242,11 +242,11 @@ export const reducer: Reducer<State> = (state: State | undefined, incomingAction
     case 'RECEIVE_REGISTRY_METADATA':
       return createState(state, {
         isLoading: false,
-        registries: {
+        registries: createState(state.registries, {
```
sleshchenko, line 245: let's rename createState to something more meaningful, to what it does, like `assign`, `fillState`

And please write a few lines of docs:
```
Fills the first (target) object with fields from the second object (source).
Note: it does not merge nested objects, maps or arrays but just overrides them.
You will need to use it on nested level where merging is needed. Like:
target = {"map": {"target": ""value", ...}
source = {"map": {"source": ""value", ...}
source.map = fillState(target.map, source.map);
fillState(target, source)
```

This PR fully should solve the referenced issue but in further PRs we may want to update `RECEIVE_DEVFILE` with the same changes.

sleshchenko (approved):

packages/dashboard-frontend/src/store/DevfileRegistries/index.ts
akurinnoy: I'm not sure that `fillState` describes what really function does. I lean to rename it to `createObject`. 

And the jsdoc will begin with:
```
 * Creates a new object where it takes the first (target) object
 * and fills it with fields from the second object(source).
 * Note: ...
```

sleshchenko: > createObject

+1

PR #315, opened 2021-08-02
Dashboard warns users that Safari is not supported 

akurinnoy:
### What does this PR do?

In this PR I added the fix that allows Dashboard to run in Safari. And since Safari is not supported users will always see the permanent banner with the warning message at the top of the page.

Screenshot 2021-08-02 at 11 40 42

fixes https://github.com/eclipse/che/issues/20123


olexii4 (approved):
LGTM

sleshchenko (approved):
LGTM

Since officially we don't support any other than Google Chrome, it makes sense to show it on the Firefox as well.

Screenshot_20210802_144045

but it out of the current PR scope. I'll create a dedicated issue and raise it on Che Community Call.

sleshchenko:
please, after merging create PR with cherry-pick the stuff onto 7.34.x

PR #321, opened 2021-08-10
fix(scm): Always get raw content from scm resolver

benoitf:
### What does this PR do?
Axios is returning a JSON object even if we specify a raw content
https://github.com/axios/axios/issues/907
Need to workaround with a custom transformResponse

Also, provide a better error message to isolate the root cause

Fixes https://github.com/eclipse/che/issues/20271


sleshchenko (commented):
LGTM

sleshchenko (approved):

akurinnoy (approved):

PR #324, opened 2021-08-26
Prevent browser to cache index.html and product.json

akurinnoy:
### What does this PR do?

This PR adds a hook for the static server that updates the `cache-control` header for resources that should not be cached by browsers.
Additionally, added redirection from `/dashboard` to `/dashboard/`.

fixes https://github.com/eclipse/che/issues/20266


akurinnoy:
[dashboard-ci-test]

PR #327, opened 2021-08-31
fix: fix devworkspace happy path by using up to date sample

sleshchenko:
### What does this PR do?
fix: fix devworkspace happy path by using up to date sample.
l0rd's sample is updated https://github.com/l0rd/spring-petclinic/pull/3 
but we depend on commit which is outdated. So, now we rely on the latest version of Che sample

No issues was created.


sleshchenko:
seems happy path test fails due Theia, but from Dashboard point of view it looks good. So, merging PR
Screenshot_20210901_115028

PR #329, opened 2021-09-01
Use @devfile/api types in the dashboard frontend

akurinnoy:
### What does this PR do?

This PR refactors the Dashboard Frontend to use types from `@devfile/api` package. It extends original interfaces provided to have some optional fields mandatory and applies them to the dashboard frontend.

fixes https://github.com/eclipse/che/issues/20191


packages/dashboard-frontend/src/pages/GetStarted/updateDevfileMetadata.ts
```diff
@@ -19,7 +19,7 @@ import {
 import getRandomString from '../../services/helpers/random';
 
 export function updateDevfileMetadata(devfile: api.che.workspace.devfile.Devfile, meta?: che.DevfileMetaData): api.che.workspace.devfile.Devfile {
-  if (isDevfileV2(devfile)) {
+  if (isDevfile(devfile)) {
```
sleshchenko, line 22: well, it seems strange... You check if devfile is devfile, for sure it is )
please consider making things clearer here

packages/dashboard-frontend/src/pages/WorkspaceDetails/Header/Actions/index.tsx
```diff
@@ -74,7 +75,8 @@ export class HeaderActionSelect extends React.PureComponent<Props, State> {
       }
       this.props.history.push(nextPath);
     } catch (e) {
-      const message = `Unable to ${selected.toLocaleLowerCase()} ${this.props.workspaceName}. ` + e.toString().replace('Error: ', '');
+      const errorMessage = getErrorMessage(e);
+      const message = `Unable to ${selected.toLocaleLowerCase()} ${this.props.workspaceName}. ` + errorMessage.replace('Error: ', '');
```
sleshchenko, line 79: before I expanded it, it `selected` seems not very clear. Please consider renaming to something like `action`, `selectedAction`, `actionToPerform`

packages/dashboard-frontend/src/pages/WorkspacesList/index.tsx
```diff
@@ -240,7 +242,8 @@ export default class WorkspacesList extends React.PureComponent<Props, State> {
       } catch (e) {
         const workspace = this.props.workspaces.find(workspace => id === workspace.id);
         const workspaceName = workspace?.name ? ` "${workspace?.name}"` : '';
-        const message = `Unable to ${workspace && isWorkspaceV1(workspace.ref) ? 'delete' : 'terminate'} workspace${workspaceName}. ` + e.toString().replace('Error: ', '');
+        const errorMessage = getErrorMessage(e);
+        const message = `Unable to ${workspace && isCheWorkspace(workspace.ref) ? 'delete' : 'terminate'} workspace${workspaceName}. ` + errorMessage.replace('Error: ', '');
```
sleshchenko, line 246:
```suggestion
        const message = `Unable to ${workspace && isCheWorkspace(workspace.ref) ? 'delete' : 'terminate'} workspace ${workspaceName}. ` + errorMessage.replace('Error: ', '');
```
?

packages/dashboard-frontend/src/services/workspace-client/devworkspace/devWorkspaceClient.ts
```diff
@@ -169,9 +168,10 @@ export class DevWorkspaceClient extends WorkspaceClient {
         sidecarPolicy,
         suffix: workspaceId,
       });
-    } catch (error) {
-      console.error(error);
-      throw new Error(`Unable to update the devWorkspace with the devfile resolver: ${error.message}`);
+    } catch (e) {
+      console.error(e);
+      const errorMessage = getErrorMessage(e);
+      throw new Error(`Unable to update the devWorkspace with the devfile resolver: ${errorMessage}`);
```
sleshchenko, line 174:
```suggestion
      throw new Error(`Unable to resolve theia plugins: ${errorMessage}`);
```

yarn.lock
```diff
@@ -16,47 +16,47 @@
   dependencies:
     "@babel/highlight" "^7.14.5"
 
-"@babel/compat-data@^7.14.5":
-  version "7.14.7"
-  resolved "https://registry.yarnpkg.com/@babel/compat-data/-/compat-data-7.14.7.tgz#7b047d7a3a89a67d2258dc61f604f098f1bc7e08"
-  integrity sha512-nS6dZaISCXJ3+518CWiBfEr//gHyMO02uDxBkXTKZDN5POruCnOZ1N4YBRZDCabwF8nZMWBpRxIicmXtBs+fvw==
+"@babel/compat-data@^7.15.0":
```
sleshchenko, line 19: please regenerate dependencies licenses checks.

sleshchenko (commented):
looked though and haven't found anything critical.
I'll take another more detailed look tomorrow.

packages/dashboard-frontend/src/pages/GetStarted/updateDevfileMetadata.ts
akurinnoy: > You check if devfile is devfile, for sure it is )

Not necessarily. 
The object named `devfile` may have a shape different from `Devfile`. This typeguard checks if the object has all required fields so it can be then used safely as `Devfile`.

I was also hesitating about this naming but decided that if this function is explicitly imported from the `devfileApi` module then it's clear enough what it is supposed to do. My assumption was wrong :)

So, let me rename this typeguard to `isDevfileV2`. And should I rename the interface in the same way `devfileApi.Devfile` to `devfileApi.DevfileV2`?

packages/dashboard-frontend/src/pages/WorkspaceDetails/Header/Actions/index.tsx
akurinnoy: yeah, it's a good time to make this clearer.

packages/dashboard-frontend/src/services/keycloak/setup.ts
```diff
@@ -134,15 +135,16 @@ export class KeycloakSetupService {
...
-      const errorMessage = 'Cannot get Keycloak settings' + (
-        e.response?.status === undefined
-          ? '. Response is not available, please check console for details.'
-          : `: ${e.response?.status} ${e.response?.statusText}`
-      );
+      let errorMessage = 'Cannot get Keycloak settings';
+      if (isAxiosError(e) && e?.response?.status) {
+        errorMessage += `: ${e.response?.status} ${e.response?.statusText}`;
+      } else {
+        errorMessage += '. Response is not available, please check console for details.';
```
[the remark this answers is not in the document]
akurinnoy: Good catch! The Network tab should be checked first of all.

yarn.lock
akurinnoy: sure!

packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
```diff
@@ -196,15 +196,15 @@ export const actionCreators: ActionCreators = {
...
     dispatch({ type: 'REQUEST_DEVWORKSPACE' });
     try {
-      let updatedWorkspace: IDevWorkspace;
+      let updatedWorkspace: devfileApi.DevWorkspace;
       if (workspace.metadata.annotations && workspace.metadata.annotations[DEVWORKSPACE_NEXT_START_ANNOTATION]) {
         // If the workspace has DEVWORKSPACE_NEXT_START_ANNOTATION then update the devworkspace with the DEVWORKSPACE_NEXT_START_ANNOTATION annotation value and then start the devworkspace
         const state = getState();
         const plugins = selectDwPluginsList(state);
-        const storedDevWorkspace = JSON.parse(workspace.metadata.annotations[DEVWORKSPACE_NEXT_START_ANNOTATION]) as IDevWorkspace;
+        const storedDevWorkspace = JSON.parse(workspace.metadata.annotations[DEVWORKSPACE_NEXT_START_ANNOTATION]) as devfileApi.DevWorkspace;
```
sleshchenko, line 207: please verify these data has corrent format

packages/dashboard-frontend/src/pages/WorkspaceDetails/Header/Actions/index.tsx
olexii4: +1 for  ```action```

packages/dashboard-frontend/webpack.config.common.js
```diff
@@ -110,7 +110,11 @@ const config = {
     ]
   },
   resolve: {
-    extensions: ['.js', '.ts', '.tsx']
+    extensions: ['.js', '.ts', '.tsx'],
+    alias: {
+      'vscode-languageserver-protocol/lib/utils/is': 'vscode-languageserver-protocol/lib/common/utils/is',
+      'vscode-languageserver-protocol/lib/main': 'vscode-languageserver-protocol/lib/node/main',
```
olexii4, line 116: Why do we need these aliases?

packages/dashboard-frontend/src/store/__mocks__/devWorkspaceBuilder.ts
```diff
@@ -10,33 +10,36 @@
...
+      uid: 'uid-' + getRandomString(4)
     },
     spec: {
       started: false,
       routingClass: 'che',
       template: {},
     },
     status: {
-      devworkspaceId: getRandomString(4),
+      devworkspaceId: 'devworkspaceId-' + getRandomString(4),
```
olexii4, line 35: I don't agree with this approach to set these values separately. It should be in two steps
The first step - should generate uid.
The second step - get devworkspaceId from uid.

https://github.com/eclipse-che/che-dashboard/pull/329/files#diff-16bcd373a2ae55c977a1ebc3e2c967d2c90cca1acbb409e1b8e7feb214e14cbdR16

```
...
  let workspaceId = 'workspace';
  if (workspace.metadata.uid) {
    workspaceId += workspace.metadata.uid.split('-').splice(0,3).join('');
  }
...
```
WDYT?

packages/dashboard-frontend/src/store/__mocks__/devWorkspaceBuilder.ts
```diff
@@ -10,33 +10,36 @@
...
       mainUrl: '',
       phase: 'STOPPED',
     }
   }
 
   withId(id: string): DevWorkspaceBuilder {
-    this.workspace.status.devworkspaceId = id;
+    this.workspace.metadata.uid = id;
```
olexii4, line 42: It is not the same 'status.devworkspaceId' and 'metadata.uid'.
We need to recover a new value for ```metadata.uid``` from the ```stratus.devworkspaceId```.

sleshchenko: well, for devworkspaces I would say we need to drop workspace id notion and use namespace/name instead.

sleshchenko: but yeah, to move in an iterative manner, it should be as Orel provided https://github.com/eclipse-che/che-dashboard/pull/329#discussion_r702012154
the original https://github.com/devfile/devworkspace-operator/blob/89eb71a53c5e61b643d93d8f07530d32df7613d9/controllers/workspace/devworkspace_controller.go

packages/dashboard-frontend/webpack.config.common.js
akurinnoy: That fixes the build after upgrading typescript.

packages/dashboard-frontend/src/containers/FactoryLoader/updateDevfileMetadata.ts
```diff
@@ -22,13 +22,13 @@ import getRandomString from '../../services/helpers/random';
 export type FactorySource = { factory?: { params: string } };
 
 export default function updateDevfileMetadata(devfile: api.che.workspace.devfile.Devfile, factoryParams: string, createPolicy: CreatePolicy): api.che.workspace.devfile.Devfile {
-  if (isDevfileV2(devfile)) {
+  if (isDevfile(devfile)) {
```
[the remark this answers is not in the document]
akurinnoy: let's fix this in a separate pull request because it demands additional refactoring

tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+    "composite": true,
+    "declaration": true,
+    "declarationMap": true,
+    "experimentalDecorators": true,
+    "emitDecoratorMetadata": true,
+    "noEmit": false,
+    "baseUrl": "./packages",
+  },
+}
+
```
sleshchenko, line 18:
```suggestion
```

sleshchenko (commented):
Nothing critical found. LGTM
I'll approve after tests are fixed and I'll test changes manually.

packages/dashboard-frontend/src/store/__mocks__/devWorkspaceBuilder.ts
```diff
@@ -10,33 +10,41 @@
...
       template: {},
     },
-    status: {
-      devworkspaceId: getRandomString(4),
-      mainUrl: '',
-      phase: 'STOPPED',
-    }
+  }
+
+  private uidToDevworkspaceId(uid: string): string {
```
olexii4, line 36: Why do we need these methods here if we are not going to use workspaces Id at all?

https://github.com/eclipse-che/che-dashboard/pull/329/files#diff-16bcd373a2ae55c977a1ebc3e2c967d2c90cca1acbb409e1b8e7feb214e14cbdR16

packages/dashboard-frontend/src/store/__mocks__/devWorkspaceBuilder.ts
```diff
@@ -10,33 +10,41 @@
...
       name: 'dev-wksp-' + getRandomString(4),
       namespace: '',
+      uid: 'uid-' + getRandomString(4)
     },
     spec: {
       started: false,
       routingClass: 'che',
       template: {},
     },
-    status: {
```
olexii4, line 31: I am sure we should leave 'status' as an empty object

```
status: { }
```

and remove things like that:
```
if (this.workspace.status === undefined) {
      this.workspace.status = this.buildStatus();
    }
```

packages/dashboard-frontend/src/store/__mocks__/devWorkspaceBuilder.ts
akurinnoy: The only purpose is to properly mock a devworkspace for the unit test.

packages/common/src/helpers/__tests__/errors.spec.ts
```diff
@@ -0,0 +1,204 @@
...
+
+  afterEach(() => {
+    jest.clearAllMocks();
+    mockAxios.resetHandlers();
+  });
+
+  it('should return default message', () => {
+    expect(getMessage(undefined)).toEqual('Unexpected error.');
+  });
+
```
olexii4, line 34: It should support the type 'string' too.

```
 it('should return error message', () => {
    const message = 'Expected error message.';
    expect(getMessage(message)).toEqual(message);
  });
```

olexii4: And have test coverage for  'isError' method:

```
  it('should check if it is an error or not', () => {
    const message = 'Expected error.';
    const error = new Error(message);
    expect(isError(error)).toEqual(true);
    expect(isError(message)).toEqual(false);
  });

```

packages/common/src/helpers/errors.ts
```diff
@@ -0,0 +1,90 @@
...
+      return `"${response.status} ${response.statusText}" returned by "${response.config.url}".`;
+    } else {
+      return `"${response.status} ${response.statusText}".`;
+    }
+  }
+
+  if (isErrorLike(error)) {
+    return error.message;
+  }
+
```
olexii4, line 56: These changes without supporting type 'string' as an Error bring a lot of wrong messages.

For example 'updateWorkspace'
The first time it is using in storage

```
  updateWorkspace: (workspace: che.Workspace): 
...
    try {
      const updatedWorkspace = await cheWorkspaceClient.restApiClient.update(workspace.id, workspace);
...
    } catch (e) {
const errorMessage = `Failed to update the workspace ${workspace.devfile.metadata.name}, reason: ` + common.helpers.errors.getMessage(e);
...
```
https://github.com/eclipse-che/che-dashboard/pull/329/files#diff-3c70d20df3faae24a13278f6cc9c832b435c183fedfeb19aea49575d56420eefR340

 The second time in the controller
```
...    
try {
      await this.props.updateWorkspace(changedWorkspace);
...
    } catch (e) {
      const errorMessage = common.helpers.errors.getMessage(e);
...

```
https://github.com/eclipse-che/che-dashboard/pull/329/files#diff-34b89a5b5c3edd3273d30d794ff491e5abd7e9c710f70b5a68a119a6383bade8R115

The resulting message always will be the same:
'Unexpected error. Check DevTools console and network tabs for more information.'

packages/common/src/helpers/errors.ts
```diff
@@ -0,0 +1,90 @@
...
+import { HttpError } from "@kubernetes/client-node";
+
+/**
+ * This helper function does its best to get an error message from the provided object.
+ * @param error An object that supposed to contain an error message, or the error message.
+ * @returns Error message
+ */
+export function getMessage(error: unknown): string {
+  if (!error) {
+    return 'Unexpected error.';
```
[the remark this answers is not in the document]
akurinnoy: I'm not sure about this suggestion. The error message is supposed to be shown to users. And I wonder what they should do when they see this message?

packages/common/src/helpers/errors.ts
akurinnoy: Nice catch! Thanks for pointing this out cause I overlooked this case.

packages/common/src/helpers/__tests__/errors.spec.ts
akurinnoy: I'll add these test cases in the following commit.

olexii4 (approved):
LGTM

packages/common/src/helpers/errors.ts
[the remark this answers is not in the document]
sleshchenko: create an issue on github with reported facing `error mustn't be undefined`. It's easier for developers to fix such a problem than `Unexpected error.`.
Maybe `error is not specified` `is empty`...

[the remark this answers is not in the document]
akurinnoy: that makes sense, I'll update the message,
thank you for elaborating on it!

sleshchenko (approved):
Tested and seems to work just fine

PR #333, opened 2021-09-04
feat: Add withCredentials option to fetch from private registry

mtsmfm:
### What does this PR do?

When registry is private, e.g. behind oauth2-proxy, it requires cookies to access.

### DIFF
--- a/packages/dashboard-frontend/src/services/registry/devfiles.ts
+++ b/packages/dashboard-frontend/src/services/registry/devfiles.ts
@@ -49,7 +49,7 @@ function resolveLinks(metadata: che.DevfileMetaData, baseUrl: string): any {
 
 export async function fetchData<T>(url: string): Promise<T> {
   try {
-    const response = await axiosInstance.get<T>(url);
+    const response = await axiosInstance.get<T>(url, {withCredentials: true});
     return response.data;
   } catch (e) {
     throw getErrorMessage(e);


packages/dashboard-frontend/src/services/registry/devfiles.ts
```diff
@@ -49,7 +49,7 @@ function resolveLinks(metadata: che.DevfileMetaData, baseUrl: string): any {
 
 export async function fetchData<T>(url: string): Promise<T> {
   try {
-    const response = await axiosInstance.get<T>(url);
+    const response = await axiosInstance.get<T>(url, {withCredentials: true});
```
sleshchenko, line 52: Sorry for the noob question: At which stage OAuth2 flow is going to be done? Or this settings allows to perform redirect and set cookies?

mtsmfm: Sorry for lacking explanation.

In my situation, both private registry and che-dashboard are placed under the same domain.

It's like the following:

- che-dashboard is served as https://che-eclipse-che.example.com
- custom private registry is served as https://custom-che-devfile-registry.example.com
- oauth2 proxy protects *.example.com and uses *.example.com cookie

In this situation, OAuth2 flow is already done because an user already visited https://che-eclipse-che.example.com.

But if an user doesn't has a valid credential for private registry even they can open che-dashboard, this API call will fail because of redirection I think.
Should we consider such a case?

mtsmfm:
I tried my fork (ghcr.io/mtsmfm/che-dashboard:pr-5) and I've noticed default devfile registry doesn't return exact hostname for Access-Control-Allow-Origin header. Probably it should return Access-Control-Allow-Credentials as well. I'll send a PR to che-devfile-registry.

packages/dashboard-frontend/src/services/registry/devfiles.ts
sleshchenko: Sounds reasonable. Thx.
Probably we need to add the same for fetching plugins for devworkspaces

sleshchenko (approved):
@akurinnoy @olexii4 please take a look as well

akurinnoy (approved):
@mtsmfm thanks for your contribution!

mtsmfm:
@sleshchenko @akurinnoy Thank you for approving but please wait until I send a patch to che-devfilre-registry.
This PR may cause error because of lacking CORS related headers.
I've changed to draft for now.

https://github.com/eclipse-che/che-dashboard/pull/333#issuecomment-913733980

olexii4 (approved):

mtsmfm:
@sleshchenko @akurinnoy @olexii4 I sent PRs:

- https://github.com/eclipse-che/che-devfile-registry/pull/466
- https://github.com/eclipse-che/che-plugin-registry/pull/1043

mtsmfm:
> @sleshchenko @akurinnoy @olexii4 I sent PRs:

Both PRs are merged, I changed state to ready for review

packages/dashboard-frontend/src/services/registry/devfiles.ts
sleshchenko: @mtsmfm I may miss some points but after I think about it a bit more, I wonder if it's secure enough to set this value without explicit configuring on CheCluster level. I just have doubt that we'll send authorization cookies in service that tend to be registry but it's not, but just there to grab user's data.

Also, are you aware of the single-host exposure strategy powered by Che Gateway?

I think the proper way would be to:
1. Run everything on the single host, which is already supported.
2. Make dashboard open a real OAuth window for user to do authentication and make sure user trust it, before sending cookies.

sleshchenko: But I may be wrong and it may be admin's responsibility to configure OpenShift OAuth or any other auth not to set *, but only domains which are expected...

mtsmfm: > are you aware of the single-host exposure strategy powered by Che Gateway?

I know the strategy but I don't prefer it cause most of services we develop are not served under sub directory.
It's preferable to be hosted similar with production I think.

sleshchenko:
It does not seem to be way to go, solution should be implemented withing https://github.com/eclipse/che/issues/20465

PR #337, opened 2021-09-08
fix: better handle K8s client error while watching

sleshchenko:
### What does this PR do?
fix: better handle K8s client error while watching.
Screenshot_20210908_140645

N/A

### DIFF
--- a/packages/dashboard-backend/src/devworkspace-client/services/api/workspace-api.ts
+++ b/packages/dashboard-backend/src/devworkspace-client/services/api/workspace-api.ts
@@ -27,7 +27,6 @@ import {
 
 import { devfileToDevWorkspace } from '../converters';
 import { helpers } from '@eclipse-che/common';
-import { HttpError } from '@kubernetes/client-node';
 
 export class DevWorkspaceApi implements IDevWorkspaceApi {
   private readonly customObjectAPI: k8s.CustomObjectsApi;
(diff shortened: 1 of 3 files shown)


olexii4 (approved):
LGTM

packages/dashboard-backend/src/devworkspace-client/services/api/workspace-api.ts
```diff
@@ -188,10 +187,19 @@ export class DevWorkspaceApi implements IDevWorkspaceApi {
       } else {
         callbacks.onError(`Error: Unknown type '${type}'.`);
       }
-    }, error => {
-      console.error('Watch in namespace error:', error);
-      const message = typeof (error) === 'string' ? error : 'Unknown type error.'
-      callbacks.onError(message);
+    }, (error: any) => {
+      let message;
```
olexii4, line 191: Could we use our ```getMessage(error: unknown): string``` here?
https://github.com/eclipse-che/che-dashboard/blob/main-7.37/packages/common/src/helpers/errors.ts#L21

sleshchenko: good idea

sleshchenko: The only problem I noticied with it:
https://github.com/eclipse-che/che-dashboard/blob/main-7.37/packages/common/src/helpers/errors.ts#L62
```
 return 'Unexpected error. Check DevTools console and network tabs for more information.'
```
which is good error for frontend but not for backend, which is my case.

leaving as is and will think how we can handle it

PR #348, opened 2021-09-16
feat: make backend respect internal URL, CAs and http proxy settings

sleshchenko:
### What does this PR do?
This PR makes dashboard backend uses internal Che Server URL when available and respect CAs, Http proxy settings in the same way as Theia does.
Probably it makes sense to move that logic to workspace-client library, but at this point I decided to take easier/quicker way since workspace-client has a lot of outdated stuff. So, I see that possible it will land to newer Che Typescript client, with devworkspaces and newer stuff.

Prereq for merging is https://github.com/eclipse-che/che-operator/pull/1073

Fixes https://github.com/eclipse/che/issues/20367

### How to test this PR?
This is copied/pasted from Che operator PR:
Http proxy and custom certs are not easy to test, so I just copied Theia solution on dashboard side and assume it works.
1. For Che Self Signed certs verification you need to test Che on minikube:
```sh
cat EOF << > /tmp/che-cluster-patch.yaml
spec:
  k8s:
    singleHostExposureType: gateway
  server:
    dashboardImage: 'sleshchenko/che-dashboard:che-ca'
    dashboardImagePullPolicy: Always
    customCheProperties:
      CHE_INFRA_KUBERNETES_ENABLE__UNSUPPORTED__K8S: "true"
EOF
chectl server:deploy --installer=operator --platform=minikube \
  --che-operator-cr-patch-yaml=che-cluster-patch.yaml \
  --che-operator-image=sleshchenko/che-operator:dashboard-ca \
  --workspace-engine=dev-workspace
```
2. Open Dashboard and make sure dashboard reports the right values in logs:
Screenshot_20210916_152206

3. Disable internal URLs
```
oc patch checluster/eclipse-che -n eclipse-che --type=merge \
    --patch "{\"spec\":{\"server\":{\"disableInternalClusterSVCNames\":true}}}"
```
4. After Che operator did its job check Che Operator and Dashboard collaborate together well, and now public Che URL is used:
Screenshot_20210916_152924

### DIFF
--- a/.deps/dev.md
+++ b/.deps/dev.md
@@ -167,6 +167,7 @@
 | [`@types/eslint-visitor-keys@1.0.0`](https://www.github.com/DefinitelyTyped/DefinitelyTyped.git) | MIT | clearlydefined |
 | [`@types/express-serve-static-core@4.17.24`](https://github.com/DefinitelyTyped/DefinitelyTyped.git) | MIT | clearlydefined |
 | [`@types/express@4.17.13`](https://github.com/DefinitelyTyped/DefinitelyTyped.git) | MIT | clearlydefined |
+| [`@types/fs-extra@9.0.12`](https://github.com/DefinitelyTyped/DefinitelyTyped.git) | MIT | clearlydefined |
 | [`@types/glob@7.1.4`](https://github.com/DefinitelyTyped/DefinitelyTyped.git) | MIT | clearlydefined |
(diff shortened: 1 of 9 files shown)


olexii4 (commented):
LGTM
I have tested it on Kubernetes Infrastructure.

olexii4 (approved):

akurinnoy (approved):

PR #323, opened 2021-08-15
feat: Make che-dashboard installable to support ctrl-w etc

mtsmfm:
### What does this PR do?

This PR makes che-dashboard PWA and installable to support shortcuts which can't be overridden e.g. ctrl-w

### What issues does this PR fix or reference?

Related to https://github.com/eclipse/che/issues/17844

#### Docs PR

----

You can try this change with ghcr.io/mtsmfm/che-dashboard:pr-5

PR #355, opened 2021-09-21
Fix web worker bundle filename, prevent cache web workers

akurinnoy:
### What does this PR do?

https://github.com/eclipse-che/che-dashboard/pull/323

This PR fixes editor web worker filename caused by resolving a merge conflict in [this PR](https://github.com/eclipse-che/che-dashboard/pull/323). Additionally, it prevents browser to cache web workers. 

N/a

### Is it tested? How?

Examine the Network tab of Developer tools to see that two web workers are loaded and none of them is cached by the browser.

### DIFF
--- a/packages/dashboard-backend/src/static.ts
+++ b/packages/dashboard-backend/src/static.ts
@@ -34,7 +34,9 @@ export function registerStaticServer(publicFolder: string, server: FastifyInstan
   const doNotCache = [
     '/dashboard/',
     '/dashboard/index.html',
-    '/dashboard/client.bundle.js',
+    '/dashboard/client.js',
+    '/dashboard/service-worker.js',
+    '/dashboard/editor.worker.js',
     '/dashboard/assets/branding/product.json',
   ];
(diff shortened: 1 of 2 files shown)


olexii4 (approved):

packages/dashboard-backend/src/static.ts
```diff
@@ -34,7 +34,9 @@ export function registerStaticServer(publicFolder: string, server: FastifyInstan
   const doNotCache = [
     '/dashboard/',
     '/dashboard/index.html',
-    '/dashboard/client.bundle.js',
+    '/dashboard/client.js',
```
olexii4, line 37: We should remove  ```'/dashboard/client.js',```  here.

olexii4: and replace ```pathData.chunk.name === 'service-worker' || 'editor.worker'``` to ```pathData.chunk.name === 'service-worker' || pathData.chunk.name === 'editor.worker'```

https://github.com/eclipse-che/che-dashboard/blob/main/packages/dashboard-frontend/webpack.config.common.js#L29

akurinnoy: you're right, fixing it

sleshchenko (approved):

PR #360, opened 2021-09-23
Upgrade devfile/api and use consts from there

sleshchenko:
### What does this PR do?
Upgrade devfile/api and use consts from there

It's a piece for https://github.com/eclipse/che/issues/20422

### Is it tested? How?
Just test devworkspace flow


dmytro-ndp:
[dashboard-ci-test]

PR #370, opened 2021-09-29
Run linters as pull request check

akurinnoy:
### What does this PR do?

Runs linters as a step of build-and-test action.

Screenshot 2021-09-30 at 10 31 13


sleshchenko (approved):
Seems good to me

olexii4 (approved):

PR #373, opened 2021-10-05
feat: Allow to switch editor on factory link URL

benoitf:
### What does this PR do?
Allow to select the editor when providing the factory URL

We can add ?`che-editor=<editor-entry-in-plugin-registry>`

https://github.com/eclipse/che/issues/20508

### Is it tested? How?
Tested by trying to select other editor from the plug-in registry
It adds `editors` entries in the dwPlugins state. Default-editor is still loaded when booting the application and we lazy load the custom editor in case it's not the same than the default one.


benoitf:
closing as it's in my fork and then tests are not launched

PR #375, opened 2021-10-05
feat: use newer theia lib

benoitf:
### What does this PR do?
Update che-theia library using devfile API
only a dependency has been updated, no code change in che-theia library logic

N/A

### Is it tested? How?
Tried to start a workspace with che-theia and it was working


sleshchenko (approved):
Thanks for your contribution. I'll merge after PR checks pass

PR #376, opened 2021-10-06
feat: DW load custom editors from factory workflow + refactor specific editor processing

benoitf:
---
***NOTE***
As there was no review on https://github.com/eclipse-che/che-dashboard/pull/374 and as I needed to move forward, I merged two PRs into a single one (with 2 commits)

---

### What does this PR do?
Allow to select the editor when providing the factory URL (in case of DevWorkspace only)

We can add ?`che-editor=<editor-entry-in-plugin-registry | url of the devfile.yaml of the editor>`

example of entries:
`eclipse/che-theia/next` (which is the default editor)
`eclipse/che-theia/latest`
`che-incubator/intellij-community/latest`

or with full URL like
`https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/che-incubator/intellij-community/latest/devfile.yaml`

Then, it will apply Theia logic only if the editor is Theia
And it removes analysis of some theia specific entries from dashboard itself, now the logic is moved the library

Editor specific logic is extracted to a dedicated class and it's easier to plug new processor.

https://github.com/eclipse/che/issues/20508
https://github.com/eclipse/che/issues/20588

### Is it tested? How?
I've tested using `http://che-host/#https://github.com/che-samples/java-spring-petclinic/tree/devfilev2?che-editor=che-incubator/intellij-community/latest` or `http://che-host/#https://github.com/che-samples/java-spring-petclinic/tree/devfilev2?che-editor=https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/che-incubator/intellij-community/latest/devfile.yaml` or also using default links like `http://che-host/#https://github.com/che-samples/java-spring-petclinic/tree/devfilev2`

Also added unit tests

### DIFF
--- a/packages/dashboard-frontend/src/services/workspace-client/devworkspace/DevWorkspaceEditorProcessTheia.ts
+++ b/packages/dashboard-frontend/src/services/workspace-client/devworkspace/DevWorkspaceEditorProcessTheia.ts
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) 2018-2021 Red Hat, Inc.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+
+import { InversifyBinding } from '@eclipse-che/che-theia-devworkspace-handler/lib/inversify/inversify-binding';
+import { CheTheiaPluginsDevfileResolver } from '@eclipse-che/che-theia-devworkspace-handler/lib/devfile/che-theia-plugins-devfile-resolver';
+import common from '@eclipse-che/common';
+import { injectable } from 'inversify';
+import { IDevWorkspaceEditorProcess, IDevWorkspaceEditorProcessingContext } from './devWorkspaceClient';
+
+/**
+ * This class manages what needs to be added on devfile/devworkspace template for Che-Theia
+ */
+@injectable()
+export class DevWorkspaceEditorProcessTheia implements IDevWorkspaceEditorProcess {
+
+    /**
+     * Match if the editor metadata name contains the name theia
+     */
+    match(context: IDevWorkspaceEditorProcessingContext): boolean {
+        return context.editorsDevfile.some(editor => editor.metadata.name.toLowerCase().includes('theia'));
+    }
+
+    /**
+     * Do the theia stuff
+     */
+    public async apply(context: IDevWorkspaceEditorProcessingContext): Promise<void> {
+        console.log('DevWorkspaceEditorProcessTheia: Applying CheTheia processor on top of the Devfile.');
+        // call theia library to insert all the logic
+        const inversifyBindings = new InversifyBinding();
+        const container = await inversifyBindings.initBindings({
+            pluginRegistryUrl: context.pluginRegistryUrl || '',
+            axiosInstance: context.axios,
+            insertTemplates: false,
+        });
+        const cheTheiaPluginsContent = context.optionalFilesContent['.che/che-theia-plugins.yaml'];
+        const vscodeExtensionsJsonContent = context.optionalFilesContent['.vscode/extensions.json'];
+
+        const cheTheiaPluginsDevfileResolver = container.get(CheTheiaPluginsDevfileResolver);
+
+        // call library to update devWorkspace and add optional templates
+        try {
+            await cheTheiaPluginsDevfileResolver.handle({
+                devfile: context.devfile,
+                cheTheiaPluginsContent,
+                vscodeExtensionsJsonContent,
+                devWorkspace: context.devWorkspace,
+                devWorkspaceTemplates: context.devWorkspaceTemplates,
+                suffix: context.workspaceId,
+            });
+        } catch (e) {
+            console.error(e);
+            const errorMessage = common.helpers.errors.getMessage(e);
+            throw new Error(`Unable to resolve theia plugins: ${errorMessage}`);
+        }
+    }
+}
--- a/packages/dashboard-frontend/src/services/workspace-client/devworkspace/devWorkspaceClient.ts
+++ b/packages/dashboard-frontend/src/services/workspace-client/devworkspace/devWorkspaceClient.ts
@@ -10,11 +10,7 @@
  *   Red Hat, Inc. - initial API and implementation
  */
 
-import { inject, injectable } from 'inversify';
-import { InversifyBinding } from '@eclipse-che/che-theia-devworkspace-handler/lib/inversify/inversify-binding';
-import { CheTheiaPluginsDevfileResolver } from '@eclipse-che/che-theia-devworkspace-handler/lib/devfile/che-theia-plugins-devfile-resolver';
-import common from '@eclipse-che/common';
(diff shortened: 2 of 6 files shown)


packages/dashboard-frontend/src/services/workspace-client/devworkspace/devWorkspaceClient.ts
```diff
@@ -80,7 +117,9 @@ export class DevWorkspaceClient extends WorkspaceClient {
   private readonly showAlert: (alert: AlertItem) => void;
 
   constructor(@inject(KeycloakSetupService) keycloakSetupService: KeycloakSetupService,
-    @inject(AppAlerts) appAlerts: AppAlerts) {
+    @inject(AppAlerts) appAlerts: AppAlerts,
+    @multiInject(IDevWorkspaceEditorProcess) private editorProcesses: IDevWorkspaceEditorProcess[],
```
olexii4, line 121: @benoitf Could you please clarify why did you use @multiInject for IDevWorkspaceEditorProcess?

Because I can find only
```
container.bind(IDevWorkspaceEditorProcess).to(DevWorkspaceEditorProcessTheia).inSingletonScope();
```
https://github.com/eclipse-che/che-dashboard/pull/376/files#diff-fb91aea76f5a40d3573cac09e17def0370d0dee56e1365729095996afa56b62eR34

benoitf: Hello @olexii4 
I use multi-inject to not tied the logic to theia directly but to 'editor processing'

As I want to plug other libraries for other editors (like VS Code / intellij) I've used multi-inject (and each editor process being responsible on saying if it's applying or not the stuff) vs having all editors being injected directly in that class

akurinnoy (approved):

olexii4 (approved):

packages/dashboard-frontend/src/store/Plugins/devWorkspacePlugins/index.ts
```diff
@@ -60,11 +87,45 @@ export type KnownAction = RequestDwPluginAction
...
+  | ReceiveDwEditorAction
+  | RequestDwEditorErrorAction;
 
 export type ActionCreators = {
   requestDwDevfiles: (url: string) => AppThunk<KnownAction, Promise<void>>;
   requestDwDefaultEditor: (settings: che.WorkspaceSettings) => AppThunk<KnownAction, Promise<void>>;
+  requestDwEditor: (settings: che.WorkspaceSettings, editorName: string) => AppThunk<KnownAction, Promise<void>>;
+}
+
+export function mergeEditors(stateActionEditorsMap: {[key: string]: PluginDefinition}, actionEditorsMap: {[key: string]: PluginDefinition}): {[key: string]: PluginDefinition} {
```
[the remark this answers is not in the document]
benoitf: yes at some point it was not the case 👍  thanks

benoitf:
removed useless mergeEditors function

packages/dashboard-frontend/src/store/Plugins/devWorkspacePlugins/index.ts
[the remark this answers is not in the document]
benoitf: I've removed the function

packages/dashboard-frontend/src/services/workspace-client/devworkspace/DevWorkspaceEditorProcessTheia.ts
```diff
@@ -0,0 +1,65 @@
...
+ * This class manages what needs to be added on devfile/devworkspace template for Che-Theia
+ */
+@injectable()
+export class DevWorkspaceEditorProcessTheia implements IDevWorkspaceEditorProcess {
+
+    /**
+     * Match if the editor metadata name contains the name theia
+     */
+    match(context: IDevWorkspaceEditorProcessingContext): boolean {
+        return context.editorsDevfile.some(editor => editor.metadata.name.toLowerCase().includes('theia'));
```
sleshchenko, line 29: that again could leave on the editor lib side but here I have a weaker feeling that it should.

packages/dashboard-frontend/src/services/workspace-client/devworkspace/devWorkspaceClient.ts
```diff
@@ -53,6 +50,46 @@ export type Subscriber = {
...
+
+  devWorkspaceTemplates: devfileApi.DevWorkspaceTemplateLike[];
+
+  editorsDevfile: devfileApi.Devfile[];
+
+  pluginRegistryUrl?: string;
+
+  axios: AxiosInstance;
+
+  optionalFilesContent: { [fileName: string]: string };
```
sleshchenko, line 72: I would replace it with a func that can provide content on-demand by doing requests to Che Server.

packages/dashboard-frontend/src/store/Plugins/devWorkspacePlugins/index.ts
```diff
@@ -94,56 +125,83 @@ export const actionCreators: ActionCreators = {
...
-      const errorMessage = 'Failed to load the default editor, reason: plugin registry URL or default editor ID is not provided by Che server.';
-      dispatch({
-        type: 'RECEIVE_DW_DEFAULT_EDITOR_ERROR',
-        error: errorMessage,
-      });
-      throw errorMessage;
-    }
+    let editorUrl;
+    // check if the editor is an id or a path to a given editor
+    if (editorName.startsWith('https')) {
```
sleshchenko, line 136:
```suggestion
    if (editorName.startsWith('https://')) {
```
could be a bit safer )

packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
```diff
@@ -347,21 +348,38 @@ export const actionCreators: ActionCreators = {
     },
     pluginRegistryUrl: string | undefined,
     pluginRegistryInternalUrl: string | undefined,
+    attributes: { [key: string]: string },
   ): AppThunk<KnownAction, Promise<void>> => async (dispatch, getState): Promise<void> => {
 
-    const state = getState();
+    let state = getState();
+
+    console.log(`Default editor is ${state.dwPlugins.defaultEditorName}`);
```
sleshchenko, line 356: maybe it makes more sense to move that log into 
```
      // take the default editor name
      cheEditor = state.dwPlugins.defaultEditorName;
```
?

packages/dashboard-frontend/src/services/workspace-client/devworkspace/devWorkspaceClient.ts
benoitf: I think it would be better to have this loading/patching stuff occurring on the server side as well.

packages/dashboard-frontend/src/store/Plugins/devWorkspacePlugins/index.ts
benoitf: agree, fixed

packages/dashboard-frontend/src/services/workspace-client/devworkspace/devWorkspaceClient.ts
benoitf: dashboard backend.
For example from client side, there is no need to download theia library or to load the default editor file
The content of the default editor should be loaded by the server side of dashboard (and the theia library as well)

PR #378, opened 2021-10-08
chore(fronend): refactor HeaderTools component

akurinnoy:
### What does this PR do?

This PR adds an application launcher into the page masthead. As for now, app launcher contains one item that allows users to open the cluster Web UI page.

resolves https://github.com/eclipse/che/issues/20106


packages/common/src/dto/application-info.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+
+export interface ApplicationInfo {
+  url: string;
+  title: string;
+  icon: string;
+  group?: string;
```
sleshchenko, line 17: what is a group?

sleshchenko: I see. Makes sense.
It seems to be consistent with what OpenShift Console does with ConsoleLink
Screenshot_20211019_172932

packages/dashboard-backend/src/devworkspace-client/services/cluster-info.ts
```diff
@@ -0,0 +1,18 @@
...
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+
+export const clusterConsoleUrl = process.env['OPENSHIFT_CONSOLE_URL'] || '';
```
olexii4, line 13: I am wondering how it is supposed to work if we do not have an environment variable OPENSHIFT_CONSOLE_URL

akurinnoy: The Applications menu won't be shown at all.

packages/dashboard-backend/src/devworkspace-client/services/cluster-info.ts
```diff
@@ -0,0 +1,18 @@
...
+export const clusterConsoleTitle = process.env['OPENSHIFT_CONSOLE_TITLE']
+  || 'Openshift Console';
+export const clusterConsoleIcon = process.env['OPENSHIFT_CONSOLE_ICON']
```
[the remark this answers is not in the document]
olexii4: WDYT?

packages/dashboard-backend/src/api/clusterInfo.ts
```diff
@@ -0,0 +1,35 @@
...
+import { clusterConsoleGroup, clusterConsoleIcon, clusterConsoleTitle, clusterConsoleUrl } from '../devworkspace-client/services/cluster-info';
+import { getSchema } from '../services/helpers';
+
+const tags = ['clusterInfo'];
+
+export function registerClusterInfo(server: FastifyInstance) {
+
+  server.get(
+    `${baseApiPath}/cluster-info`,
+    getSchema({ tags, params: clusterInfoSchema }),
```
olexii4, line 26: It is not the right schema for request
```
{
  type: 'object',
  properties: {
    title: { type: 'string' },
    url: { type: 'string' },
    icon: { type: 'string' },
  }
}
```

akurinnoy: Any hints on how to fix that?

akurinnoy: I need to get rid of clusterInfoSchema because it describes the request but not the response.

PR #380, opened 2021-10-12
Dashboard reports its version

akurinnoy:
### What does this PR do?

In the About window

- renamed "Version" to "Server Version"
- added "Dashboard Version" section

Screenshot 2021-10-12 at 12 34 03

fixes https://github.com/eclipse/che/issues/20474


packages/dashboard-frontend/src/Layout/Header/Tools/about-modal.tsx
```diff
@@ -48,9 +49,17 @@ const AboutModalItems: React.FC<AboutModalItemsProps> = (
...
+              <TextListItem component='dt'>Dashboard Version</TextListItem>
+              <TextListItem component='dd'>
+                <div className='co-select-to-copy'>{dashboardVersion}</div>
+              </TextListItem>
+            </>
+          )}
           {productVersion && (
             <>
-              <TextListItem component='dt'>Version</TextListItem>
+              <TextListItem component='dt'>Product Version</TextListItem>
```
sleshchenko, line 62: Maybe it's better to report it as `Server Version` since it's version of Che/CRW Server.
@nickboldt WDYT?

nickboldt: Yes but in like 3 months "server" will be "devworkspace engine" ...

What's wrong with Product Version?

sleshchenko: Yes and no. We still will have Che Server for some duties. But its exact version may be not critical, so I'm OK with leaving Product as well.

About DevWorkspace operator version, we can try to report DWO CSV version, but in addition.

packages/dashboard-frontend/src/Layout/Header/Tools/__tests__/about-modal.spec.tsx
```diff
@@ -55,9 +55,20 @@ describe('About modal', () => {
...
+    };
+    const { getByText } = render(component);
+    expect(getByText('Product Version')).not.toBeNull();
+    expect(getByText('1.2.3')).not.toBeNull();
+  });
+
   it('should display product version', () => {
     const { getByText } = render(component);
-    expect(getByText('Version')).not.toBeNull();
+    expect(getByText('Product Version')).not.toBeNull();
```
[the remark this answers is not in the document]
sleshchenko: It won't be full replacement but if you think product makes more sense, I'm OK with that

olexii4 (approved):

PR #387, opened 2021-10-21
Convert devfile v2 to devfile v1

akurinnoy:
### What does this PR do?

In case if devworkspaces mode is disabled devfiles v2 should be converted to devfiles v1.

dashboard part for https://github.com/eclipse/che/issues/20644

### Is it tested? How?

1. Deploy Che with the following CheCluster CR fragment:
```yaml
spec:
  server:
    cheImage: quay.io/skabashn/che-server
    # is built from https://github.com/eclipse-che/che-server/pull/168
    cheImageTag: main_2021_10_19_09_45_15 
    cheImagePullPolicy: Always
    dashboardImage: quay.io/eclipse/che-dashboard:pr-387
    dashboardImagePullPolicy: Always
```     

2. Create a workspace using either factory flow or import from git with this URL: `https://github.com/che-samples/java-spring-petclinic/tree/devfilev2`

### DIFF
--- a/packages/dashboard-frontend/src/containers/FactoryLoader/devfile-converter.ts
+++ b/packages/dashboard-frontend/src/containers/FactoryLoader/devfile-converter.ts
@@ -25,8 +25,6 @@ import devfileApi from '../../services/devfileApi';
 
 export class DevfileConverter {
 
-  private INFRASTRUCTURE_NAMESPACE = 'infrastructureNamespace';
-
   componentEndpointV2toComponentEndpointV1(
     componentEndpoints?: V220DevfileComponentsItemsContainerEndpoints[]
   ): cheApi.workspace.devfile.Endpoint[] | undefined {
@@ -254,10 +252,6 @@ export class DevfileConverter {
           // skip attribute that value is not a string
           if (typeof devfileV2.metadata.attributes?.[attributeName] !== 'string') {
             return;
-          }
-          if (attributeName !== this.INFRASTRUCTURE_NAMESPACE) {
-            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-            attributes[attributeName] = devfileV2.metadata.attributes![attributeName];
           }
         });
         if (Object.keys(attributes).length > 0) {


packages/dashboard-frontend/src/store/FactoryResolver/devfile-service.ts
```diff
@@ -0,0 +1,203 @@
...
+  cpuRequest?: string;
+  dedicatedPod?: boolean;
+  endpoints?: DevfileComponentEndpoint[];
+  env?: DevfileComponentEnv[];
+  image: string;
+  memoryLimit?: string;
+  memoryRequest?: string;
+  mountSources?: boolean;
+  sourceMapping?: string;
+  volumeMounts?: DevfileComponentVolumeMount[];
```
sleshchenko, line 144: well, these model interfaces should be dropped I believe since now we have up to date/correct model from @devfile/api

akurinnoy: good point! I'll fix that

packages/dashboard-frontend/src/store/FactoryResolver/index.ts
```diff
@@ -130,7 +138,12 @@ export const actionCreators: ActionCreators = {
       if (cheEditor) {
         optionalFilesContent['.che/che-editor.yaml'] = cheEditor;
       }
-      const devfile = getDevfile(data, location);
+      let devfile = getDevfile(data, location);
+
+      if (isDevfileV2(devfile) && selectCheDevworkspaceEnabled(state) === false) {
+        devfile = devfileConverter.devfileV2toDevfileV1(devfile);
```
sleshchenko, line 144: we would need to override the message to make it clear that devfile v2 is converted to v1
Screenshot_20211021_144450

Like `devfile 2.x version is found. Converted it to version 1`

packages/dashboard-frontend/src/store/FactoryResolver/devfile-converter.ts
```diff
@@ -0,0 +1,280 @@
...
+    if (devfileV2.metadata.attributes) {
+      const attributeKeys = Object.keys(devfileV2.metadata.attributes);
+      if (attributeKeys.length > 0) {
+        const attributes = devfileV1.attributes || {};
+        attributeKeys.forEach(attributeName => {
+          // skip attribute that value is not a string
+          if (typeof devfileV2.metadata.attributes?.[attributeName] !== 'string') {
+            return;
+          }
+          if (attributeName !== this.INFRASTRUCTURE_NAMESPACE) {
```
sleshchenko, line 251: it can be dropped since it must not be defined at devfile v2 level

PR #396, opened 2021-11-08
fix: complete support of che-editor.yaml file

benoitf:
### What does this PR do?
Handle id, reference, inline and override properties

Fixes https://github.com/eclipse/che/issues/20715

### Is it tested? How?
Tested with default samples and in addition:
+ https://github.com/benoitf/override-editor-memory (id definition + override)
+ https://github.com/benoitf/use-reference-editor (external definition + override)
+  https://github.com/benoitf/exvscd (ide is not displayed directly but we can see that it's using the inline definition)

### DIFF
--- a/packages/dashboard-frontend/src/services/workspace-client/devworkspace/devWorkspaceClient.ts
+++ b/packages/dashboard-frontend/src/services/workspace-client/devworkspace/devWorkspaceClient.ts
@@ -34,6 +34,7 @@ import { WorkspaceAdapter } from '../../workspace-adapter';
 import { safeLoad } from 'js-yaml';
 import { DEVWORKSPACE_UPDATING_TIMESTAMP_ANNOTATION } from '../../devfileApi/devWorkspace/metadata';
 import { AxiosInstance } from 'axios';
+import { V220DevfileComponentsItemsContainer } from '@devfile/api';
 
 export interface IStatusUpdate {
   error?: string;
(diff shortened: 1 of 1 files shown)


benoitf:
it would be nice to have it included in the upcoming tag 7.39

packages/dashboard-frontend/src/services/workspace-client/devworkspace/devWorkspaceClient.ts
```diff
@@ -260,18 +274,77 @@ export class DevWorkspaceClient extends WorkspaceClient {
...
+        } else {
+          repositoryEditorYamlUrl = `${pluginRegistryUrl}/plugins/${cheEditorYaml.id}/devfile.yaml`;
+        }
+      } else if (cheEditorYaml.reference) {
+        // load the content of this editor
+        console.debug(`Loading editor from reference ${cheEditorYaml.reference}`);
+        repositoryEditorYamlUrl = cheEditorYaml.reference;
+      }
+      if (repositoryEditorYamlUrl) {
+        const response = await this.axios.get<string>(repositoryEditorYamlUrl, {
```
olexii4, line 311: It will cause 'has been blocked by CORS policy' error as we see with your [example](https://raw.githubusercontent.com/benoitf/use-reference-editor/main/.che/che-editor.yaml):
```
...
reference: https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/eclipse/che-theia/next/devfile.yaml
...
```

It should be done on the dashboard-backend

.

olexii4 (approved):
I like this proposal. I have tested it. It works except CORS error which happens in the case with external reference (https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/eclipse/che-theia/next/devfile.yaml).

I agree to merge it if we open an issue to add tests for this important functionality (because we already have one regression https://github.com/eclipse/che/issues/20737 with P1 and we should prevent it in the future).

packages/dashboard-frontend/src/services/workspace-client/devworkspace/devWorkspaceClient.ts
benoitf: @olexii4 it's strange, github.io is allowing CORS
requests are returning `access-control-allow-origin: *` and it was working on my side

benoitf:
@olexii4 it's not already handled by https://github.com/eclipse/che/issues/20733 ?

akurinnoy (approved):

PR #403, opened 2021-11-18
Provide default devfile v2 in the devworkspace mode

akurinnoy:
### What does this PR do?

When Che is running devworkspace mode, the dashboard should supply the default devfile v2 instead of a devfile v1 resolved by Che Server.

fixes https://github.com/eclipse/che/issues/20705

### Is it tested? How?

Create a workspace using factory flow from `<che-host>#https://github.com/eclipse/che`

### DIFF
--- a/packages/dashboard-frontend/src/services/helpers/__tests__/getDefaultDevfile.spec.ts
+++ b/packages/dashboard-frontend/src/services/helpers/__tests__/getDefaultDevfile.spec.ts
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) 2018-2021 Red Hat, Inc.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+
+import getDefaultDevfile from '../getDefaultDevfile';
+
+describe('getDefaultDevfile', () => {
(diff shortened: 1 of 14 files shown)


packages/dashboard-frontend/src/services/helpers/__tests__/getDefaultDevfile.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+
+import getDefaultDevfile from '../getDefaultDevfile';
+
+describe('getDefaultDevfile', () => {
+  it('should return devfile v1', () => {
+    const isDevworkspacesEnabled = true;
```
olexii4, line 17: It should be for devfile v2

olexii4 (approved):

packages/dashboard-frontend/src/store/FactoryResolver/getDevfile.ts
```diff
@@ -27,69 +29,88 @@ import {
 export function getDevfile(
   data: FactoryResolver,
   location: string,
-): api.che.workspace.devfile.Devfile {
+  isDevworkspacesEnabled: boolean,
```
olexii4, line 32: This param should be added to the description too:
```
/**
 * Returns a devfile from the FactoryResolver object.
 * @param data a FactoryResolver object
 * @param location a source location
 ```
 =>
 ```
/**
 * Returns a devfile from the FactoryResolver object.
 * @param data a FactoryResolver object
 * @param location a source location
  * @param isDevworkspacesEnabled ...
 ```

PR #410, opened 2021-11-25
fix: handle links that may not be string but objects

benoitf:
### What does this PR do?
Handle cases where links are not only with string type

Fixes https://github.com/eclipse/che/issues/20827

### Is it tested? How?
Click on any 'sample'

### DIFF
--- a/packages/dashboard-frontend/src/services/registry/devfiles.spec.ts
+++ b/packages/dashboard-frontend/src/services/registry/devfiles.spec.ts
@@ -43,7 +43,6 @@ describe('devfile links', () => {
     const object = '/devfile/foo.yaml';
     const updated = updateObjectLinks(object, baseUrl);
 
-    // this one is not updated as already absolute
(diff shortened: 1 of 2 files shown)


packages/dashboard-frontend/src/services/registry/devfiles.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+    expect(resolved.devWorkspaces['eclipse/che-theia/next']).toBe(
+      `${baseUrl}/devfiles/nodejs-react/devworkspace-che-theia-next.yaml`,
+    );
+  });
+
+  it('should update links', () => {
+    const object = '/devfile/foo.yaml';
+    const updated = updateObjectLinks(object, baseUrl);
+
+    // this one is not updated as already absolute
```
akurinnoy, line 46:
```suggestion
```

akurinnoy (approved):

olexii4 (approved):

PR #424, opened 2021-12-13
fix: request devfile JSON schema

akurinnoy:
### What does this PR do?

This PR fixes the issue that the dashboard is not able ot fetch a devfile schema.

fixes https://github.com/eclipse/che/issues/20919

### Is it tested? How?

Use the dashboard image from this PR


benoitf (approved):
Tested using the image and I don't reproduce the error

Also I'm able to start DevWorkspaces

ibuziuk (approved):

PR #429, opened 2021-12-22
Create devworkspaces using generated DevWorkspace and DevWorkspaceTemplate

akurinnoy:
### What does this PR do?

This PR adds the ability to create devworkspaces using pre-generated resources. 
When a user clicks on a sample, the dashboard analyzes the sample looking for devWorkspace links. Among those links it takes one for the default editor, adds that as a query parameter to factory link, and runs factory flow. The Factory Loader detects that query param and skips steps for resolving and fetching devfiles and processing devworkspace templates to apply editor-specific stuff. Instead, it fetches pre-generated resources, creates devworkspace template and devworkspace, and starts the devworkspace.

<details>
<summary>Screencast</summary>

https://user-images.githubusercontent.com/16220722/147071281-7e25f92a-eef3-461e-a46e-2c3864fbe027.mov

</details>

resolves https://github.com/eclipse/che/issues/20829

### Is it tested? How?

Create a workspace from the Quick Add page.


dmytro-ndp:
[dashboard-ci-test]

packages/dashboard-frontend/src/containers/FactoryLoader/index.tsx
```diff
@@ -458,6 +463,67 @@ export class FactoryLoaderContainer extends React.PureComponent<Props, State> {
...
+        const resources = safeLoadAll(yamlContent);
+        const devworkspace = resources.find(
+          resource => resource.kind === 'DevWorkspace',
+        ) as devfileApi.DevWorkspace;
+        const devworkspaceTemplate = resources.find(
+          resource => resource.kind === 'DevWorkspaceTemplate',
+        ) as devfileApi.DevWorkspaceTemplate;
+
+        await this.props.createWorkspaceFromPrebuiltResources(devworkspace, devworkspaceTemplate);
+
```
olexii4, line 511: With this solution, we will be able to create only one workspace from a sample.

But, as a developer, I prefer to have 'a quick start' flow, when I can choose a sample with needed technology,  create a new workspace. Then, add a new project inside, push it to upstream without requirement to recreate a new workspace. 

Note: 
We  need to recreate workspace, because with this implementation click on a sample creates a new devworkspace  from prebuilt resources without changing its name(adding some generated symbols as it was previosly). 
And we can't create two workspaces with different name from the same sample, can't rename existing devworkspace. So, it will require to recreate each workspace wich was based on that sample.

My proposal: improve this implementation and add some generated symbols to devworkspace and devworkspacetemplate names

WDYT?

packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
```diff
@@ -411,6 +415,37 @@ export const actionCreators: ActionCreators = {
       }
     },
 
+  createWorkspaceFromPrebuiltResources:
```
olexii4, line 418:
```suggestion
  createWorkspaceFromResources:
```

packages/dashboard-frontend/src/services/workspace-client/devworkspace/devWorkspaceClient.ts
```diff
@@ -239,7 +239,28 @@ export class DevWorkspaceClient extends WorkspaceClient {
     return workspace;
   }
 
-  async create(
+  async createFromPrebuiltResources(
```
olexii4, line 242:
```suggestion
  async createFromResources(
```

packages/dashboard-frontend/src/containers/FactoryLoader/index.tsx
akurinnoy: Good point, thanks!
Actually, I don't have any strong opinion on how do we proceed with this. 
@ibuziuk @l0rd can you please share your thoughts?

ibuziuk: > With this solution, we will be able to create only one workspace from a sample.

I believe this is fine for the first iteration. Basically,  if a workspace based on the sample already exists, clicking on the devfile sample again will reopen the existing one. For the time being I would avoid doing any modifications of the DevWorkspace / DevWorkspace template on UD level

olexii4 (approved):

PR #431, opened 2021-12-24
[feat] Implement editor agnostic kubeconfig injection

JPinkney:

### What does this PR do?
This PR implements editor agnostic kubeconfig injection for devworkspaces by calling a new endpoint `/namespace/:namespace/devworkspaceId/:devworkspaceId/kubeconfig` everytime a devworkspace is loaded. This new API injects the kubeconfig into all containers associated with a devworkspaceId.

https://github.com/eclipse/che/issues/20782

### Is it tested? How?

Ensure that Che is running with devworkspaces enabled then either create a new devworkspace from the dashboard itself or use `https://gist.githubusercontent.com/JPinkney/d4f6e665d1239a5fc4caefbe20bf4f56/raw/96faebbd51b3bba209f8a8de8b448327f385c779/a.yaml` with a factory url (that gist has vscode-kubernetes-tools and vscode-yaml pre-installed). Wait for the workspace to start and then observe that the kubeconfig is located at `$HOME/.kube/config`. If you are using the gist with vscode-kubernetes-tools then you can also observe that the kubeconfig is automatically loaded into vscode-kubernetes-tools when you open up the side panel.

### DIFF
--- a/packages/dashboard-backend/src/api/kubeConfigAPI.ts
+++ b/packages/dashboard-backend/src/api/kubeConfigAPI.ts
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2018-2021 Red Hat, Inc.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+
+import { FastifyInstance, FastifyReply, FastifyRequest } from 'fastify';
+import { baseApiPath } from '../constants/config';
+import { getDevWorkspaceClient } from './helper';
+import { getSchema } from '../services/helpers';
+import { restParams } from '../typings/models';
+
+const tags = ['kubeconfig'];
+
+export function registerKubeConfigApi(server: FastifyInstance) {
+  server.post(
+    `${baseApiPath}/namespace/:namespace/devworkspaceId/:devworkspaceId/kubeconfig`,
+    getSchema({ tags }),
+    async function (request: FastifyRequest, reply: FastifyReply) {
+      try {
+        const { kubeConfigApi } = await getDevWorkspaceClient(request);
+        const { namespace, devworkspaceId } = request.params as restParams.INamespacedPodParam;
+        await kubeConfigApi.injectKubeConfig(namespace, devworkspaceId);
+        reply.code(204);
+        return reply.send();
+      } catch (e) {
+        return reply.send(e);
+      }
+    },
+  );
+}
--- a/packages/dashboard-backend/src/devworkspace-client/index.ts
+++ b/packages/dashboard-backend/src/devworkspace-client/index.ts
@@ -17,6 +17,7 @@ import {
   IDevWorkspaceClient,
   IDevWorkspaceTemplateApi,
   IDockerConfigApi,
+  IKubeConfigApi,
 } from './types';
 import { findApi } from './services/helpers';
 import { DevWorkspaceTemplateApi } from './services/api/template-api';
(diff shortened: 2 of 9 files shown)


packages/dashboard-frontend/src/containers/IdeLoader.tsx
```diff
@@ -422,6 +423,13 @@ class IdeLoaderContainer extends React.PureComponent<Props, State> {
     }
     const workspace = this.props.allWorkspaces.find(workspace => workspace.id === cheWorkspace.id);
     if (workspace && workspace.ideUrl) {
+      try {
+        if (workspace.isDevWorkspace) {
+          await injectKubeConfig(workspace.namespace, workspace.id);
```
olexii4, line 428: @JPinkney I have several questions:
1. Why should we do it each time when we open IDE?
This solution is going to do it each time when open running IDE from the Dashboard 

2. Could we move this functionality on the server side? I am not sure that it is the right place to do it (client-side of UD).

@akurinnoy WDYT?

akurinnoy: @olexii4 I have the same thoughts about it.

JPinkney: > 1. Why should we do it each time when we open IDE?
> This solution is going to do it each time when open running IDE from the Dashboard

If the home directory is not persisted then the kubeconfig would only be injected the first time and then never injected ever again. It calls injectKubeConfig each time but if a kubeconfig has already been injected its essentially a no-op.

> 2. Is it possible to move this part to the devworkspace-operator side to inject kubeconfig there?

The devworkspace operator has no way of figuring out who the user is or what their token is that's why it was decided to put this mechanism on the dashboard. Similiar-ish to how kubeconfig injection works with Web Terminal Operator except this solution is more generic, works will all editors and doesn't require che-machine-exec. Other then that the mechanism of action is the exact same as how the OpenShift Console does it.

> 2a. Could we move this functionality to the server-side at least?

We _might_ be able to technically move it to the server side but what's the gain? We still need something to initiate the call to the backend when the pod is fully started or just constantly wait until the pod status is started.

olexii4: If the home directory is not persistent, we need to inject the kube-config on the workplace start each time. 

This solution does not work for such actions as Run Workspace in background and Restart Workspace. This can lead to errors in the case of single-host.

I would suggest transferring this logic to the startWorkspace action (devworkspaces redux store) to have this implementation in a single place.

JPinkney: > If the home directory is not persistent, we need to inject the kube-config on the workplace start each time.

That's how its currently implemented

> This solution does not work for such actions as Run Workspace in background and Restart Workspace. This can lead to errors in the case of single-host.

Can you elaborate? Why wouldn't it work?

> I would suggest transferring this logic to the startWorkspace action (devworkspaces redux store) to have this implementation in a single place.

You can't inject the kubeconfig until all the pods are ready. At best I might be able to watch the pod from the startWorkspace action and inject after its in a running state

olexii4: > You can't inject the kubeconfig until all the pods are ready. At best I might be able to watch the pod from the startWorkspace action and inject after its in a running state

It is the best solution(watch the pod from the startWorkspace action and inject after it's in a running state).

We have an example here:
https://github.com/eclipse-che/che-dashboard/blob/main/packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts#L315-L336

olexii4: > Can you elaborate? Why wouldn't it work?

Use single-host mode. Imagine two tabs that will be open in the browser(workspaces list and workspace editor). Click on Restart action.

olexii4: "Watch the pod from the startWorkspace action" should fix it.

olexii4 (approved):
LGTM

akurinnoy (approved):

PR #440, opened 2022-01-17
Add owner references into devWorkspace templates

olexii4:
### What does this PR do?
Add owner references templates and source information into devWorkspace

fixes https://github.com/eclipse/che/issues/21013
fixes https://github.com/eclipse/che/issues/21022

### Is it tested? How?

1. Click on a sample to create a workspace
2. Delete the workspace
3. Try to create a workspace from the previous sample
4. Click on the same sample one more time - the existing workspace would be opened

### DIFF
--- a/packages/dashboard-frontend/src/containers/__tests__/FactoryLoader.spec.tsx
+++ b/packages/dashboard-frontend/src/containers/__tests__/FactoryLoader.spec.tsx
@@ -33,6 +33,7 @@ import {
   actionCreators as factoryResolverActionCreators,
   isOAuthResponse,
 } from '../../store/FactoryResolver';
+import getRandomString from '../../services/helpers/random';
 
 const showAlertMock = jest.fn();
 const setWorkspaceQualifiedName = jest.fn();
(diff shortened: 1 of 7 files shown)


packages/dashboard-frontend/src/containers/__tests__/FactoryLoader.spec.tsx
```diff
@@ -181,6 +181,78 @@ describe('Factory Loader container', () => {
...
+
+      render(
+        <Provider store={store}>
+          <FactoryLoaderContainer {...props} />
+        </Provider>,
+      );
+
+      const elementHasError = screen.getByTestId('factory-loader-has-error');
+      await waitFor(() => expect(elementHasError.innerHTML).toEqual('true'));
+      await waitFor(() => expect(createWorkspaceFromResourcesMock).not.toHaveBeenCalled());
```
akurinnoy, line 228: it makes sense to wait for the mock to be called but not vice versa
```suggestion
      expect(createWorkspaceFromResourcesMock).not.toHaveBeenCalled();
```

olexii4: It makes sense. Thanks. I have fixed it.

akurinnoy (approved):

PR #443, opened 2022-01-20
fix: fix update to next version

mkuznyetsov:

### What does this PR do?
Instead of `SNAPSHOT`, use `next` prefix, to signify development version in `main` branch
Fix updating the next version, as PR may not get greated due to a specific error message

https://github.com/eclipse/che/issues/21044

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@eclipse-che/common",
-  "version": "7.43.0-SNAPSHOT",
+  "version": "7.43.0-next",
   "repository": "https://github.com/eclipse-che/che-dashboard",
   "license": "EPL-2.0",
   "private": true,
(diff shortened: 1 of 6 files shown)


packages/common/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@eclipse-che/common",
-  "version": "7.37.0-SNAPSHOT",
+  "version": "7.43.0-SNAPSHOT",
```
akurinnoy, line 3: shouldn't be this version changed to `x.y.z-next` as well?

nickboldt (approved):
not sure we need the -next suffix... but otherwise I like this.

For the .x branches, will they also have -next?

mkuznyetsov:
@nickboldt next suffix could help to differentiate between tags and non-tags ...if that's needed ofc

mkuznyetsov:
@nickboldt -next suffix in bugfix branch may simply provide a hint, that this code (or in the resulting artifact image) was built from the branch, and not tag. And that the next tag made from this branch would be the current version, just without suffix
if that makes sense

PR #448, opened 2022-01-27
Fix local start flow

olexii4:
### What does this PR do?
This PR fixes local start flow with OpenShift and upgrade vulnerable libraries.

### Is it tested? How?

```
# 1. Install all dependencies:
yarn
# 2. Patch cluster to enable local development flow:
yarn start:prepare
# 3. Run server locally:
yarn start
# 4. (optional) Patch cluster to revert it to initial state:
yarn start:revert
# If you want to make sure the latest bits are used, add flag to recompile
# yarn start --force-build
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,37 +49,6 @@ yarn start:revert
 
 The development server serves the dashboard-frontend and dashboard-backend on [http://localhost:8080](http://localhost:8080).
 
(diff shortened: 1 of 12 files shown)


ibuziuk (approved):
verified against dogfooding

image

akurinnoy (approved):

PR #453, opened 2022-02-04
 Provide devfile parent section to a new devworkspace

akurinnoy:
### What does this PR do?

With this fix, the dashboard will not ignore `parent` section of a devfile while creating new workspaces.

https://github.com/eclipse/che/issues/21098


olexii4 (approved):
LGTM

ibuziuk (approved):
Great job!
maybe it would be a good idea for adding test cases for all types of parent references - uri, kubernetes

akurinnoy:
@ibuziuk 

> maybe it would be a good idea for adding test cases for all types of parent references - uri, kubernetes

I'm not sure we need to do that because the dashboard simply propagates what the devfile parent section contains. Unless the dashboard should validate that content.

PR #454, opened 2022-02-04
fix: generateName is not executed if devfile v2

nils-mosbach:
See discussion here: https://github.com/eclipse/che/issues/21067

---

If starting a devfile v2 that has generateName instead of name v2 checks fail since they only check for `name`. See: https://github.com/nils-mosbach/che-dashboard/blob/f9c2a7c087871c6095b033729c51729f779f252c/packages/dashboard-frontend/src/services/devfileApi/typeguards.ts#L19-L24

This leads to the following message:

``` 
Failed to create a workspace. Failed to create a new workspace from the devfile, reason: Failed to create a new workspace. Unable to create devworkspace: DevWorkspace.workspace.devfile.io "devfile-test" is invalid: [<nil>: Invalid value: "": "spec.template.components" must validate one and only one schema (oneOf). Found none valid, spec.template.components.container: Required value, spec.template.components.name: Invalid value: "": spec.template.components.name in body should match '^[a-z0-9]([-a-z0-9]*[a-z0-9])?$']
```

This pull request executes generateName to name conversion on all Devfiles (not just v1).

### DIFF
--- a/packages/dashboard-frontend/src/containers/FactoryLoader/index.tsx
+++ b/packages/dashboard-frontend/src/containers/FactoryLoader/index.tsx
@@ -423,11 +423,11 @@ export class FactoryLoaderContainer extends React.PureComponent<Props, State> {
       if (isCheDevfile(devfile)) {
         attrs.stackName = stackName;
         delete attrs.factoryParams;
-        if (!devfile.metadata.name && devfile.metadata.generateName) {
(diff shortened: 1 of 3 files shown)


akurinnoy (approved):

ibuziuk (approved):
thank you for the contribution :+1:

PR #458, opened 2022-02-08
fix: skip DevWorkspaceTemplate updates for sidecarless editors (che-code)

nils-mosbach:
After Che-Dashboard is loaded DevWorkspaceTemplates are automatically updated to their latest plugins definition. This breaks existing workspaces for editors (e.g. che-code) that use sidecarless mode.

While starting new workspaces Che-Dashboard will remove editor definitions from DevWorkspaceTemplate since these get merged to the users dev container. If DevWorkspaceTemplate get updated, DevWorkspaceTemplate and DevWorkspace both contain editor definitions. Workspace therefore will fail to start due to duplicate ports.

Quick fix is skipping the update in case of `che-code`. Maybe `intellij` and `pycharm` would also make sense.

See also: https://github.com/eclipse/che/issues/21085

### DIFF
--- a/packages/dashboard-frontend/src/services/workspace-client/devworkspace/devWorkspaceClient.ts
+++ b/packages/dashboard-frontend/src/services/workspace-client/devworkspace/devWorkspaceClient.ts
@@ -927,6 +927,11 @@ export class DevWorkspaceClient extends WorkspaceClient {
         pluginsByUrl[url] = plugin;
       }
 
+      if (plugin.metadata.name.toLowerCase().includes('che-code')) {
+        // skip updates for editors in sidecarless mode
+        continue;
+      }
+
       const spec: Partial<V1alpha2DevWorkspaceTemplateSpec> = {};
       for (const key in plugin) {
         if (key !== 'schemaVersion' && key !== 'metadata') {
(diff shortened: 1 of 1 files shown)


nils-mosbach:
I'm happy to add IntelliJ-Editors to the condition as well, but didn't want to add these without feedback.

packages/dashboard-frontend/src/services/workspace-client/devworkspace/devWorkspaceClient.ts
```diff
@@ -927,6 +927,12 @@ export class DevWorkspaceClient extends WorkspaceClient {
         pluginsByUrl[url] = plugin;
       }
 
+      const skipEditors = ['che-code', 'idea', 'pycharm'];
```
nils-mosbach, line 930: I've added `idea` and `pycharm` to the list. (sidecarless)

benoitf (approved):

ibuziuk (approved):

olexii4:
@nils-mosbach @benoitf wouldn't it be better to prevent updating DevWorkspaceTemplates from the very beginning? I think about removing this annotation for the specified editors:

```
[COMPONENT_UPDATE_POLICY]: 'managed'
```

see https://github.com/eclipse-che/che-dashboard/blob/main/packages/dashboard-frontend/src/services/workspace-client/devworkspace/devWorkspaceClient.ts#L464

olexii4:
@nils-mosbach That attributes are used only by Che-Dashboard.

nils-mosbach:
@olexii4 done. Thanks for the hint.

akurinnoy (approved):

olexii4:
@nils-mosbach Thank you for your contribution

PR #461, opened 2022-02-17
Add annotation with workspace ID

akurinnoy:
### What does this PR do?

Each new workspace created via converting an old Che7 workspace will have a new annotation (`controller.devfile.io/devworkspace_id_override`) added so Che7 workspace folder can be mounted and reused after migrating to DevWorkspace.
Additionally, the dashboard hides the "Delete" button for those old workspaces that were converted. This should prevent accidentally deleting a Che7 workspace with all related resources while they are in use by a new devworkspace.

fix https://github.com/eclipse/che/issues/21153

### Is it tested? How?

unit tests


olexii4 (approved):
LGTM

ibuziuk (approved):

PR #470, opened 2022-03-09
feat: do not allow starting a DevWorkspace if there is already one running

ibuziuk:
### What does this PR do?
Do not allow starting a DevWorkspace if there is already one running

https://github.com/eclipse/che/issues/21032

Based on the internal discussion, for the time being, it was decided not to add extra property similar to `CHE_LIMITS_USER_WORKSPACES_RUN_COUNT` since we only support the common PVC strategy, so by definition only a single running workspace is allowed for RWO volumes.

### Is it tested? How?
do not  allow more workspaces

verbose

#### Docs PR
N/A


ibuziuk:
closing in favour of https://github.com/eclipse-che/che-dashboard/pull/472

PR #472, opened 2022-03-09
feat: do not allow starting a DevWorkspace if there is already one running

ibuziuk:
### What does this PR do?
Do not allow starting a DevWorkspace if there is already one running

https://github.com/eclipse/che/issues/21032

Based on the internal discussion, for the time being, it was decided not to add extra property similar to `CHE_LIMITS_USER_WORKSPACES_RUN_COUNT` since we only support the common PVC strategy, so by definition only a single running workspace is allowed for RWO volumes.

### Is it tested? How?

against the dogfooding instance:

do not  allow more workspaces

verbose

#### Docs PR
N/A

### DIFF
--- a/packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
+++ b/packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
@@ -262,6 +262,13 @@ export const actionCreators: ActionCreators = {
     async (dispatch, getState): Promise<void> => {
       dispatch({ type: 'REQUEST_DEVWORKSPACE' });
       try {
+        const { workspaces } = await devWorkspaceClient.getAllWorkspaces(
+          workspace.metadata.namespace,
+        );
+        const runningWorkspaces = workspaces.filter(workspace => workspace.spec.started === true);
+        if (runningWorkspaces.length > 0) {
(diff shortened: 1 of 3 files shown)


packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
```diff
@@ -262,6 +262,13 @@ export const actionCreators: ActionCreators = {
     async (dispatch, getState): Promise<void> => {
       dispatch({ type: 'REQUEST_DEVWORKSPACE' });
       try {
+        const defaultKubernetesNamespace = selectDefaultNamespace(getState());
+        const defaultNamespace = defaultKubernetesNamespace.name;
+        const { workspaces } = await devWorkspaceClient.getAllWorkspaces(defaultNamespace);
```
amisevsk, line 267:
```suggestion
        const { workspaces } = await devWorkspaceClient.getAllWorkspaces(workspace.metadata.namespace);
```

amisevsk (approved):
I'm still getting familiar with this repo, but LGTM

amisevsk:
Looks like there's a formatting error to be fixed

akurinnoy (approved):

PR #477, opened 2022-03-14
Fix unit tests

akurinnoy:
### What does this PR do?

This PR fixes unit tests for the IdeLoader container in order to make sure that defined mocks are not called.


amisevsk (commented):
Too late to approve but :+1: anyways -- thanks!

PR #480, opened 2022-03-17
Bump up nodejs version and upgrade dependencies

akurinnoy:
### What does this PR do?

1. Increases minimal supported version of nodeJS in `package.json`
2. Updates Dockerfiles to use nodeJS v16 images
3. Upgrades dependencies packages within the specified range

fixes https://github.com/eclipse/che/issues/21237


packages/common/package.json
```diff
@@ -7,7 +7,7 @@
   "main": "lib/index.js",
   "engines": {
     "yarn": ">=1.20.0",
-    "node": ">=12"
```
olexii4, line 10: Do we really need it?

packages/dashboard-frontend/package.json
```diff
@@ -16,7 +16,7 @@
   "license": "EPL-2.0",
   "engines": {
     "yarn": ">=1.20.0",
-    "node": ">=12"
```
olexii4, line 19: Do we really need it?

package.json
```diff
@@ -12,7 +12,7 @@
   "homepage": "https://github.com/eclipse-che/che-dashboard",
   "engines": {
     "yarn": ">=1.20.0",
-    "node": ">=12"
```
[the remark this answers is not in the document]
akurinnoy: Agree, I added the fixup

olexii4 (approved):
Good job!

ibuziuk (approved):
:+1: looks like one conflict on the rhel dockerfile needs to be resolved

akurinnoy:
@nickboldt please note that I updated `rhel.Dockerfile`. Since nodeJS v12 is reaching EOL I changed images to `registry.access.redhat.com/ubi8/nodejs-16:1-18`. Will that work for you?

 https://github.com/eclipse-che/che-dashboard/pull/480/files#diff-d62ecfc7db2707ad7af6609a692e3b0237d5b063661a2fac1dd115b2a63ea1c8R11
https://github.com/eclipse-che/che-dashboard/pull/480/files#diff-d62ecfc7db2707ad7af6609a692e3b0237d5b063661a2fac1dd115b2a63ea1c8R36

.github/workflows/pr.yml
```diff
@@ -43,7 +43,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        node-version: [12.x, 14.x]
+        node-version: [14.x, 16.x]
```
nickboldt, line 46: if we've updated to build with 16, why also use 14 here?

package.json
```diff
@@ -12,7 +12,7 @@
...
+    "node": ">=14"
```
nickboldt, line 15: why not `>=16`?

nickboldt (approved):
generally OK with this, but not sure why wer're doing anything for Node 14 support. Why not just Node 16 minimum?

.github/workflows/pr.yml
akurinnoy: Took this from our chatter:
The reason is to minimize switching environments forth and back to different types of versions while working on a few projects and switching between them.

package.json
akurinnoy: answered https://github.com/eclipse-che/che-dashboard/pull/480#discussion_r833380337

nickboldt:
> build-and-test (12.x) 

Should this PR check also be updated?

PR #519, opened 2022-04-29
Fix preload-script build

olexii4:

### What does this PR do?
Fix preload-script build.


ibuziuk:
@olexii4 thanks, please backport to 7.46.x once merged :+1:

ibuziuk (approved):

l0rd:
@olexii4 can you please share some details about the bug that is fixed by this PR? This is to understand if that's worth talking about it in the release notes or not.

olexii4:
> @olexii4 can you please share some details about the bug that is fixed by this PR? This is to understand if that's worth talking about it in the release notes or not.

@l0rd It was a Webpack build bug. We started to reuse the same code for building a separate preload script(which is built separately) and inside the main dashboard bundle. Webpack optimizes the build and tries to reuse part of the code from the separate bundle. As result, Webpack optimization brings the error because of routing.
We missed this regression first because of Browser Caching.

With those changes, I added a separate implementation that will be used for building preload script only.

PR #551, opened 2022-06-03
fix: polyfill crypto package

dkwon17:
Signed-off-by: David Kwon <dakwon@redhat.com>

### What does this PR do?
Since webpack v5 doesn't automatically polyfill node packages (see this [doc](https://gist.github.com/ef4/d2cf5672a93cf241fd47c020b9b3066a#webpack-5-node-polyfills-upgrade-cheatsheet)), this PR introduces polyfill for the Crypto package.

This PR prevents this error message from happening on devworkspace startup:
image

### Is it tested? How?

Check out this PR and follow [these steps](https://github.com/eclipse-che/che-dashboard#running-locally-against-remote-che-cluster) to run the dashboard locally against a cluster with Che + Devworkspace engine enabled.

Create a new devworkspace from the dashboard. The devworkspace should start successfully without the error message in the screenshot above.

### DIFF
--- a/packages/dashboard-frontend/webpack.config.common.js
+++ b/packages/dashboard-frontend/webpack.config.common.js
@@ -137,13 +137,14 @@ const config = {
       "fs": false,
       "net": false,
       "module": false,
-      "crypto": false,
       "path": false,
       "os": false,
+      "crypto": require.resolve("crypto-browserify"),
+      "stream": require.resolve("stream-browserify"),
     },
   },
   resolveLoader: {},
-  node: false,
+  node: { global: true },
   plugins: [
     new webpack.ProvidePlugin({
       Buffer: ['buffer', 'Buffer'],
--- a/packages/dashboard-frontend/package.json
+++ b/packages/dashboard-frontend/package.json
@@ -43,6 +43,7 @@
     "axios": "^0.21.1",
     "buffer": "^6.0.3",
     "connected-react-router": "^6.7.0",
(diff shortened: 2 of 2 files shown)


packages/dashboard-frontend/webpack.config.common.js
```diff
@@ -137,13 +137,14 @@ const config = {
...
-      "crypto": false,
       "path": false,
       "os": false,
+      "crypto": require.resolve("crypto-browserify"),
+      "stream": require.resolve("stream-browserify"),
     },
   },
   resolveLoader: {},
-  node: false,
+  node: { global: true },
```
dkwon17, line 147: I have added this to fix a `global is not defined` error in the front end.

dkwon17:
Alternatively, instead of providing polyfills, this PR fixes the issue by adding the crypto-js npm package instead: https://github.com/eclipse-che/che-dashboard/compare/main...dkwon17:fix-crypto?expand=1

Both solutions seem to work, but are there any preference as to solution is more preferred?

olexii4 (approved):
LGTM

ibuziuk (approved):

olexii4:
As for me, I would prefer a solution with ```crypto-js```.
@akurinnoy WDYT?

akurinnoy (approved):

akurinnoy:
@olexii4 I don't have a strong opinion but this implementation looks preferable to me.

olexii4 (approved):

PR #557, opened 2022-06-10
fix: allow devfiles without `metadata.name`

akurinnoy:
### What does this PR do?

This PR allows the dashboard to create workspaces from devfiles without the `metadata` section.

fixes https://github.com/eclipse/che/issues/21429

### Is it tested? How?

Open link https://eclipse-che.apps.cluster-wb6n9.wb6n9.sandbox506.opentlc.com/#https://github.com/l0rd/tilt-example-java/tree/cc88c7f683e5ab50d36722e7b8366d871f6a73d4

### DIFF
--- a/packages/dashboard-backend/src/devworkspace-client/services/api/workspace-api.ts
+++ b/packages/dashboard-backend/src/devworkspace-client/services/api/workspace-api.ts
@@ -66,16 +66,21 @@ export class DevWorkspaceApi implements IDevWorkspaceApi {
     }
   }
 
-  async create(devworkspace: V1alpha2DevWorkspace): Promise<V1alpha2DevWorkspace> {
+  async create(
+    devworkspace: V1alpha2DevWorkspace,
+    namespace: string,
+  ): Promise<V1alpha2DevWorkspace> {
     try {
-      if (!devworkspace.metadata?.name || !devworkspace.metadata?.namespace) {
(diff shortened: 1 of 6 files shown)


packages/dashboard-backend/src/devworkspace-client/services/api/workspace-api.ts
```diff
@@ -68,8 +68,8 @@ export class DevWorkspaceApi implements IDevWorkspaceApi {
 
   async create(devworkspace: V1alpha2DevWorkspace): Promise<V1alpha2DevWorkspace> {
     try {
-      if (!devworkspace.metadata?.name || !devworkspace.metadata?.namespace) {
-        throw 'DevWorkspace.metadata with name and namespace are required';
+      if (!devworkspace.metadata?.namespace) {
+        throw 'DevWorkspace `metadata.namespace` is required';
       }
```
amisevsk, line 73: A safer check might be that either `generateName` or `name` is present, as if both are missing the object will be rejected during creation.

akurinnoy: Thanks! Fixed!

amisevsk (approved):

ibuziuk (approved):

olexii4 (approved):
LGTM

PR #562, opened 2022-06-13
Fix kube config flow

olexii4:

### What does this PR do?
Fix kube config flow

fixes https://github.com/eclipse/che/issues/21454

### Is it tested? How?

1. Deploy Eclipse CHE
2. Create a new workspace from one of the samples. 
3.  `config` file should be in `~/.kube/` 
```
$ ls ~/.kube/
```

You can test it with https://eclipse-che.apps.cluster-k865s.k865s.sandbox811.opentlc.com

### DIFF
--- a/packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
+++ b/packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
@@ -730,39 +730,42 @@ async function onStatusUpdateReceived(
     workspaceId?: string;
   },
 ) {
-  const { status, namespace, workspaceId } = statusUpdate;
-  if (status !== statusUpdate.prevStatus) {
-    dispatch({
-      type: 'UPDATE_DEVWORKSPACE_STATUS',
-      workspaceUID: statusUpdate.workspaceUID,
-      message: statusUpdate.message,
-      status,
-    });
-    if (status === DevWorkspaceStatus.RUNNING && namespace && workspaceId) {
-      injectKubeConfig(namespace, workspaceId).catch(e => console.error(e));
+  const { status, message, prevStatus, workspaceUID, namespace, workspaceId } = statusUpdate;
+
+  if (status !== prevStatus) {
+    const type = 'UPDATE_DEVWORKSPACE_STATUS';
+    dispatch({ type, workspaceUID, message, status });
+
+    const onChangeCallback = onStatusChangeCallbacks.get(workspaceUID);
+    if (onChangeCallback) {
+      onChangeCallback(status);
     }
   }
 
-  const callback = onStatusChangeCallbacks.get(statusUpdate.workspaceUID);
-  if (callback && status) {
-    callback(status);
+  if (message && status !== DevWorkspaceStatus.STOPPED) {
+    let log: string;
+    if (status === DevWorkspaceStatus.FAILED || status === DevWorkspaceStatus.FAILING) {
+      log = `1 error occurred: ${message}`;
+    } else {
+      log = message;
+    }
+
+    dispatch({
+      type: 'UPDATE_DEVWORKSPACE_LOGS',
+      workspacesLogs: new Map([[workspaceUID, [log]]]),
+    });
   }
 
-  if (statusUpdate.message) {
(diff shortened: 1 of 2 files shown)


packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
```diff
@@ -715,20 +725,25 @@ export const reducer: Reducer<State> = (state: State | undefined, action: KnownA
...
+  const { status, namespace, workspaceId } = statusUpdate;
   if (status !== statusUpdate.prevStatus) {
     dispatch({
       type: 'UPDATE_DEVWORKSPACE_STATUS',
       workspaceUID: statusUpdate.workspaceUID,
       message: statusUpdate.message,
       status,
     });
+    if (status === DevWorkspaceStatus.RUNNING && namespace && workspaceId) {
+      injectKubeConfig(namespace, workspaceId).catch(e => console.error(e));
```
akurinnoy, line 742: I'd ask you not to mix async-await and promises chaining if that's possible. In this case, I expected to see the `try-catch` block surrounding the promise and missed the `.catch()` method attached. Such things make it harder to review code now and support it later.

```suggestion
      try {
        injectKubeConfig(namespace, workspaceId);
      } catch (e) {
        console.error(e);
      }
```

akurinnoy (changes requested):

ibuziuk (commented):
tested locally against dogfooding and there is one thing I can not grok:
```
bash-4.4$ cat ~/.kube/config 
{"apiVersion":"v1","kind":"Config","clusters":[{"name":"api-che-dev-x6e0-p1-openshiftapps-com:6443","cluster":{"server":"https://api.che-dev.x6e0.p1.openshiftapps.com:6443","insecure-skip-tls-verify":false}}],"users":[{"name":"developer","user":{"token":"<token>"}}],"contexts":[{"name":"logged-user","context":{"user":"developer","cluster":"api-che-dev-x6e0-p1-openshiftapps-com:6443","name":"logged-user"}}],"preferences":{},"current-context":"logged-user"}
bash-4.4$ oc whoami
ibuziuk
```
Basically, `oc whoami` reports the right user, but the kube config file refers to the `developer` user. wondering if this is expected? Basically, the token that is injected in file is the correct one and belongs to my user, but it is a bit strange to see the `developer` reference.

ibuziuk:
@akurinnoy thanks for the pointer https://github.com/eclipse-che/che-dashboard/blob/main/packages/dashboard-backend/src/services/kubeclient/kubeConfigProvider.ts#L39-L46

looks like the logic for retrieving the username is failing atm. Can be fixed in a separate PR though

packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
olexii4: I don't need an additional delay in this part of the code(injectKubeConfig is an expensive operation). If you have a proposal on how to do it without additional delays, I will be glad to see another solution.

But, as I see, the current solution
```
injectKubeConfig(namespace, workspaceId).catch(e => console.error(e));
```
 is faster. I don't see the reason to add additional delay.

akurinnoy: I see your point, thanks. Still, I don't think it's a bottleneck that requires us to do this.

Another solution would be not to throw an error in the `injectKubeConfig` method. Will it work for you?

ibuziuk (approved):
LGTM :+1: 
the problem with username detection can be fixed in a separate PR I believe. I think we should not fall back on token itself for username since it is a sha256 one-way hash that can not be decrypted.

akurinnoy (approved):

packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
akurinnoy: thanks!

l0rd:
Thanks @olexii4 for the rapid fix. I am not sure if it's even necessary to get the right username. It looks like the only important thing is the token anyway.

PR #589, opened 2022-07-22
Fix rules for redirecting to the sign-in page

olexii4:

### What does this PR do?
This PR fixed rules for redirecting to the sign-in page.

In the case of response error 403, we had a problem with signing redirection. 
It could happen depending on the platform setup:
Знімок екрана 2022-07-22 о 13 41 57
```...DevWorkspace storage-type attribute cannot be changed once the workspace has been created.```
fixes https://github.com/eclipse/che/issues/21568

### CODE REVIEW

File: packages/dashboard-frontend/src/services/workspace-client/index.ts [hunk 1]
```diff
@@ -11,67 +11,63 @@
...
 import { signIn } from '../helpers/login';
+import common from '@eclipse-che/common';
 
 /**
  * This class manages the common functions between the che workspace client and the devworkspace client
  */
 @injectable()
 export abstract class WorkspaceClient {
+  private loginCheckerAxiosInstance: AxiosInstance;
+  private static loginCheckerTimeout: number | undefined;
```
akurinnoy, line 27:
```suggestion
  private sessionCheckAxiosInstance: AxiosInstance;
  private static sessionCheckTimeout: number | undefined;
```

akurinnoy (approved):

ibuziuk (approved):

PR #602, opened 2022-08-09
Added the possibility to create an empty workspace

olexii4:

### What does this PR do?
Added the possibility to create an empty workspace

fixes https://github.com/eclipse/che/issues/21554

depends on https://github.com/eclipse-che/che-operator/pull/1486

### Is it tested? How?

1. Deploy CHE and open Create Workspace tab
2. Click on the Empty Workspace sample

The PR can be tested here: https://eclipse-che.apps.cluster-pdfv6.pdfv6.sandbox1498.opentlc.com/dashboard/

Знімок екрана 2022-08-17 о 16 06 23


packages/dashboard-frontend/static/registry/devfiles/empty.yaml
```diff
@@ -0,0 +1,8 @@
+schemaVersion: 2.1.0
+metadata:
+  generateName: playground
```
ibuziuk, line 3: let's call it empty to avoid confusion ;-)

ibuziuk: also, why can't we simply have it as:

```
schemaVersion: 2.2.0
metadata:
  generateName: empty
```
is it related to the https://github.com/eclipse/che/issues/21560? 

e.g. this minimal devfile would fail if we do not have the default components set in the CR e.g. 
```
  defaultComponents:
      - container:
          image: 'quay.io/devfile/universal-developer-image:ubi8-latest'
          sourceMapping: /projects
        name: universal-developer-image
```

ibuziuk, line 3:
```suggestion
  generateName: empty
```

olexii4: I have fixed it

l0rd:
I have tested this PR and it worked just fine. A few minor comments that can be addressed later:

- The `community` label should be removed as it's not sourced from a `community` devfile registry
- The description just says "Empty workspace" (as the title) and something like "Start an empty remote development environment and create files or clone a git repository afterwards" would be better
- Instead of having one unique "Empty workspace" tile it would be better to have one tile per supported editor. For example for VS Code:
   - Title: "VS Code remote dev environment"
   - Description: "Start an empty VS Code based development environment and create files or clone a git repository afterwards"
   - Icon: VS Code icon 	
- Since those are special dev environemnts, I would have them on a dedicated row. In other words after the empty dev environments we should add a new line.

ibuziuk (approved):
Great job! verified locally against dogfooding 
works like a charm :+1:

akurinnoy (approved):

nickboldt:
This is also in Dev Spaces 3.3.

PR #607, opened 2022-08-15
Provide links to [stop/go to] running workspace when trying to start a workspace when workspace limit reached

dkwon17:
Signed-off-by: dkwon17 <dakwon@redhat.com>

### What does this PR do?
When starting a workspace when there is already a workspace running, since the running workspace limit is `1` by default, there are now new links to:
* go to the running workspace
* stop the running workspace and restart the current workspace

image

When clicking "Close running workspace (bash) and restart golang-example", if there is an error when trying to stop the workspace, and alert with an error message will be displayed on the top right:

image

When there is more than one running workspace, there is a 'Return to Dashboard' link instead:

image

<details>
<summary>Demo video</summary>

https://user-images.githubusercontent.com/83611742/184721559-4265d656-b07e-4140-99e8-20a69d6e8e0a.mov

</details>

https://github.com/eclipse/che/issues/21359

### Is it tested? How?

1. Check out this PR and run locally: See [readme](https://github.com/eclipse-che/che-dashboard#running-locally-against-remote-che-cluster-nodejs-v16) (or, test with this docker image: `quay.io/eclipse/che-dashboard:pr-607`)
2. Start any workspace
3. Start another workspace and verify that you see the following:
image

4. Verify that both links work
6. Start a second workspace (ie. set a DevWorkspace's `spec.started` to `true`)
7. Start another workspace from the dashboard and verify that the dashboard prevents you from starting a workspace, and that you see the "Return to dashboard" link 
8. Verify that the "Return to dashboard" link works

To test the alert that appears on the top right of the dashboard when there is an error when stopping a workspace, I've created a dashboard image: `quay.io/dkwon17/che-dashboard:multiworkspace-improvement-error` that implements the following change:
```diff
diff --git a/packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts b/packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
index 5cfea278..ad2d8b3b 100644
--- a/packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
+++ b/packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
@@ -417,6 +417,7 @@ export const actionCreators: ActionCreators = {
     (workspace: devfileApi.DevWorkspace): AppThunk<KnownAction, Promise<void>> =>
     async (dispatch): Promise<void> => {
       try {
+        throw '<error message here>';
         await devWorkspaceClient.changeWorkspaceStatus(workspace, false);
         dispatch({
           type: 'DELETE_DEVWORKSPACE_LOGS',
```
1. Run the dashboard with the following image: `quay.io/dkwon17/che-dashboard:multiworkspace-improvement-error`
2. Start any workspace
3. Start another workspace and verify that the dashboard prevents you from starting the workspace and that you see the `Close running workspace and restart [current workspace]` link.
4. Click on the link, verify that an error alert appears on the top right.

### DIFF
--- a/packages/dashboard-frontend/src/containers/Loader/AbstractStep.tsx
+++ b/packages/dashboard-frontend/src/containers/Loader/AbstractStep.tsx
@@ -24,7 +24,11 @@ export type LoaderStepProps = {
   onRestart: () => void;
 };
 export type LoaderStepState = {
-  lastError?: string;
+  lastError?: LoaderStepError;
+};
+export type LoaderStepError = {
+  message: string;
+  error: unknown;
 };
 export abstract class AbstractLoaderStep<
   P extends LoaderStepProps,
@@ -61,7 +65,10 @@ export abstract class AbstractLoaderStep<
     const currentStep = loaderSteps.get(currentStepIndex).value;
 
     currentStep.hasError = true;
-    const lastError = common.helpers.errors.getMessage(e);
+    const lastError = {
+      message: common.helpers.errors.getMessage(e),
+      error: e,
+    };
     this.setState({
       lastError,
     });
--- a/packages/dashboard-frontend/src/containers/Loader/AbstractStep.tsx
+++ b/packages/dashboard-frontend/src/containers/Loader/AbstractStep.tsx
@@ -10,7 +10,6 @@
  *   Red Hat, Inc. - initial API and implementation
  */
 
-import common from '@eclipse-che/common';
 import React from 'react';
 import { Cancellation, pseudoCancellable } from 'real-cancellable-promise';
 import { List, LoaderStep } from '../../components/Loader/Step';
@@ -24,12 +23,9 @@ export type LoaderStepProps = {
   onRestart: () => void;
 };
 export type LoaderStepState = {
-  lastError?: LoaderStepError;
+  lastError?: unknown;
 };
-export type LoaderStepError = {
-  message: string;
-  error: unknown;
-};
+
 export abstract class AbstractLoaderStep<
   P extends LoaderStepProps,
   S extends LoaderStepState,
@@ -65,12 +61,8 @@ export abstract class AbstractLoaderStep<
     const currentStep = loaderSteps.get(currentStepIndex).value;
 
     currentStep.hasError = true;
-    const lastError = {
-      message: common.helpers.errors.getMessage(e),
-      error: e,
-    };
     this.setState({
-      lastError,
+      lastError: e,
     });
   }
 
--- a/packages/dashboard-frontend/src/containers/Loader/Workspace/Steps/StartWorkspace/index.tsx
+++ b/packages/dashboard-frontend/src/containers/Loader/Workspace/Steps/StartWorkspace/index.tsx
@@ -17,7 +17,7 @@ import common from '@eclipse-che/common';
 import { AppState } from '../../../../../store';
 import { selectAllWorkspaces, selectLogs } from '../../../../../store/Workspaces/selectors';
 import * as WorkspaceStore from '../../../../../store/Workspaces';
-import { WorkspaceLoaderPage } from '../../../../../pages/Loader/Workspace';
+import WorkspaceLoaderPage from '../../../../../pages/Loader/Workspace';
 import { DevWorkspaceStatus } from '../../../../../services/helpers/types';
 import { DisposableCollection } from '../../../../../services/helpers/disposable';
 import { delay } from '../../../../../services/helpers/delay';
@@ -77,7 +77,7 @@ class StepStartWorkspace extends AbstractLoaderStep<Props, State> {
       return true;
     }
     // set the error for the current step
-    if (this.state.lastError !== nextState.lastError) {
+    if (this.state.lastError?.message !== nextState.lastError?.message) {
       return true;
     }
     return false;
@@ -171,7 +171,14 @@ class StepStartWorkspace extends AbstractLoaderStep<Props, State> {
         // do not switch to the next step
         return false;
       } catch (e) {
-        throw new Error(common.helpers.errors.getMessage(e));
+        const message = common.helpers.errors.getMessage(e);
+
+        if (common.helpers.errors.isError(e)) {
+          // throw original error
+          e.message = message;
+          throw e;
+        }
+        throw new Error(message);
       }
     }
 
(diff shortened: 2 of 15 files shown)


packages/dashboard-frontend/src/containers/Loader/Workspace/Steps/StartWorkspace/index.tsx
```diff
@@ -171,7 +171,14 @@ class StepStartWorkspace extends AbstractLoaderStep<Props, State> {
         // do not switch to the next step
         return false;
       } catch (e) {
-        throw new Error(common.helpers.errors.getMessage(e));
+        const message = common.helpers.errors.getMessage(e);
+
+        if (common.helpers.errors.isError(e)) {
+          // throw original error
+          e.message = message;
```
olexii4, line 178: I don't think changing the original error message is a good idea.

```
throw new Error(message);
```
Is enough for each case

packages/dashboard-frontend/src/containers/Loader/AbstractStep.tsx
```diff
@@ -61,7 +65,10 @@ export abstract class AbstractLoaderStep<
     const currentStep = loaderSteps.get(currentStepIndex).value;
 
     currentStep.hasError = true;
-    const lastError = common.helpers.errors.getMessage(e);
+    const lastError = {
```
olexii4, line 68: @dkwon17 As for me, I prefer to have the lastError as a simple error or as a message. Do you have a reason to have both?

@akurinnoy WDYT?

dkwon17: On second thought, it looks like I don't need both. Is there a preference as to which to keep, which to remove?

If lastError is a message (string), then we don't need [RunningWorkspaceesExceededError](https://github.com/eclipse-che/che-dashboard/blob/53fc942af03fbbc11be8a08de5f1707cb8bfe162/packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts#L56-L61), and can instead do a string comparison for the error message [here](https://github.com/eclipse-che/che-dashboard/blob/53fc942af03fbbc11be8a08de5f1707cb8bfe162/packages/dashboard-frontend/src/pages/Loader/Workspace/index.tsx#L67)

dkwon17: In my latest push, I have `lastError` as a simple error (`unknown` type) for now:
https://github.com/eclipse-che/che-dashboard/blob/7aecab30a8fcf108bf959e9471c7860a0527a103/packages/dashboard-frontend/src/containers/Loader/AbstractStep.tsx#L26

packages/dashboard-frontend/src/containers/Loader/Workspace/Steps/StartWorkspace/index.tsx
dkwon17: I notice that if I throw a new error:
```
throw new Error(message);
```
, then this comparison [here](https://github.com/eclipse-che/che-dashboard/blob/53fc942af03fbbc11be8a08de5f1707cb8bfe162/packages/dashboard-frontend/src/pages/Loader/Workspace/index.tsx#L67) will always be false, because the original error (`RunningWorkspacesExceeded` error thrown [here](https://github.com/eclipse-che/che-dashboard/blob/53fc942af03fbbc11be8a08de5f1707cb8bfe162/packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts#L298)) is lost

In my new commit, I set it as:
https://github.com/eclipse-che/che-dashboard/blob/7aecab30a8fcf108bf959e9471c7860a0527a103/packages/dashboard-frontend/src/containers/Loader/Workspace/Steps/StartWorkspace/index.tsx#L175-L178

WDYT?

dkwon17:
My latest push also adds an alert on the top right corner if there was an error with the
```
Close running workspace (bash) and restart golang-example
```
link:
image

dkwon17:
This PR is now ready for review, I've updated the original PR description with steps to test this PR

packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
```diff
@@ -408,6 +417,7 @@ export const actionCreators: ActionCreators = {
     (workspace: devfileApi.DevWorkspace): AppThunk<KnownAction, Promise<void>> =>
     async (dispatch): Promise<void> => {
       try {
+        throw '<error message here>';
```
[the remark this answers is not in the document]
dkwon17: Oops, yes, removed

akurinnoy (approved):

ibuziuk (approved):

dkwon17:
Latest commit updates the link order and link text:
image

packages/dashboard-frontend/src/containers/Loader/Workspace/Steps/StartWorkspace/index.tsx
olexii4: @dkwon17  I think this problem should be solved there

https://github.com/eclipse-che/che-dashboard/blob/main/packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts#L353-L363

```
        throw errorMessage;
```
->
```
        throw new Error(errorMessage);
```
WDYT?

nickboldt:
My only concern here (and maybe you've accounted for this already) is...

what if the user is allowed to run more than one concurrent workspace? Do you...

* check for the # of concurrent running workspaces before redirecting to this message page? (eg., if 1 open and limit is 5, just open the 2nd workspace)
* include a list of multiple open workspaces to close/switch to, or 
* grab the first one and provide choices for closing/switching to that first one, ignoring the others?

dkwon17:
@nickboldt in that case the first case happens:

> * check for the # of concurrent running workspaces before redirecting to this message page? (eg., if 1 open and limit is 5, just open the 2nd workspace)

The only time the new links introduced in this PR appears is when the runningLimit has been reached.

If the limit is more than 1, and the limit has been reached, at the moment this PR doesn't introduce a list of multiple open workspaces to close/switch to, but just shows a "Return to dashboard" link instead:

<details>
<summary>Screenshot</summary>

image

</details>

But the list idea does sound like a very nice enhancement

nickboldt (approved):
> Return to dashboard

That works for this PR! +1, good enough. 

> But the list idea does sound like a very nice enhancement

Feel free to open a separate GH issue to track / prioritize that enhancement. Thanks!

(Note that I'm not a TS expert so my review is based on screenshots and descriptions, not a technical code review.)

olexii4 (approved):

dkwon17:
Thank you everyone for the feedback 👍 , I plan to merge this PR tomorrow

@nickboldt I made an issue with the enhancement: https://github.com/eclipse/che/issues/21660

PR #609, opened 2022-08-19
Provides OpenVSX CheCluster URL to user containers

olexii4:
### What does this PR do?
Provides OpenVSX CheCluster URL to user containers

fixes https://github.com/eclipse/che/issues/21563

### Is it tested? How?

Знімок екрана 2022-08-23 о 17 09 09

### DIFF
--- a/packages/dashboard-frontend/src/services/workspace-client/devworkspace/devWorkspaceClient.ts
+++ b/packages/dashboard-frontend/src/services/workspace-client/devworkspace/devWorkspaceClient.ts
@@ -138,6 +138,7 @@ export class DevWorkspaceClient extends WorkspaceClient {
   private readonly maxStatusAttempts: number;
   private readonly pluginRegistryUrlEnvName: string;
   private readonly pluginRegistryInternalUrlEnvName: string;
+  private readonly openVSXUrlEnvName: string;
   private readonly dashboardUrlEnvName: string;
   private readonly websocketClient: WebsocketClient;
   private webSocketEventEmitter: EventEmitter;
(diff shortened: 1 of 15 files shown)


packages/dashboard-frontend/src/services/workspace-client/devworkspace/devWorkspaceClient.ts
```diff
@@ -537,22 +551,33 @@ export class DevWorkspaceClient extends WorkspaceClient {
...
+      }
+      if (pluginRegistryInternalUrl) {
+        envs.push({
           name: this.pluginRegistryInternalUrlEnvName,
           value: pluginRegistryInternalUrl,
-        },
-      );
+        });
+      }
+      if (openVSXUrl) {
```
benoitf, line 575: Could you provide an empty environment variable

because on CheCode we need to know if it's empty or not

if it's missing, we don't know if it's empty or if we're running on previous che versions that are not providing the value ?

benoitf: so we should always have the env variable defined, just that it can contains a value or be empty

olexii4: I have fixed it

olexii4: Знімок екрана 2022-08-22 о 17 00 52

benoitf: thanks @olexii4

benoitf (approved):
LGTM

ibuziuk (approved):

benoitf:
Hi, will it be merged soon ?

akurinnoy (approved):

PR #641, opened 2022-10-03
Add unit tests to the backend package

akurinnoy:
### What does this PR do?

This PR adds unit tests for the dashboard backend.


packages/dashboard-backend/jest.config.js
```diff
@@ -19,10 +19,11 @@ module.exports = {
...
-    'src/api/**/*.ts',
-    'src/services/**/*.ts',
-    'src/devworkspace-client/*.ts',
-    'src/devworkspace-client/services/**/*.ts',
+  coveragePathIgnorePatterns: [
+    'src/devworkspaceClient/__tests__/*',
   ],
+  testPathIgnorePatterns: [
+    'src/devworkspaceClient/__tests__/errors.spec.ts',
+    'src/devworkspaceClient/__tests__/integration.spec.ts',
```
olexii4, line 27: @akurinnoy I think we should remove  'src/devworkspaceClient/__tests__/errors.spec.ts' and
    'src/devworkspaceClient/__tests__/integration.spec.ts' at all.

packages/dashboard-backend/src/__tests__/app.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+
+import { FastifyInstance } from 'fastify';
+import { setup, teardown } from '../helpers/tests/appBuilder';
+
+describe('App', () => {
```
olexii4, line 16: I’m not sure I get what this test does.

akurinnoy: It tests that the app can start. I'll update the test to make that clear.

packages/dashboard-backend/src/devworkspaceClient/const/__tests__/models.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+
+import { deletePolicy, deletionOptions } from '../models';
+
+test('models', () => {
```
olexii4, line 15: I think we don't use those models in code anymore. It should be removed.

olexii4 (approved):

PR #657, opened 2022-10-26
Remove dependency on /user and /profile API

akurinnoy:
### What does this PR do?

This PR removes the dependency on `/user` and `/profile` che-server API. Instead, the dashboard backend gets a new endpoint `/userprofile` that allows retrieving the email and username. Additionally, the Account page gets temporarily hidden as it will be reworked in further PRs.

resolves https://github.com/eclipse/che/issues/21595


packages/dashboard-frontend/src/Layout/Header/Tools/UserMenu/__tests__/index.spec.tsx
```diff
@@ -44,27 +44,14 @@ describe('User Menu', () => {
...
+  const store = createStore(username, email);
   const history = createHashHistory();
-  const user = {
-    id: 'test-id',
-    name: name,
-    email: email,
-    links: [],
-  };
   const branding = selectBranding(store.getState());
-  const userProfile = selectUserProfile(store.getState());
```
olexii4, line 57: Could you remove all unused imports after your changes too?

olexii4 (approved):
LGTM

packages/dashboard-frontend/src/Layout/Header/Tools/UserMenu/__tests__/index.spec.tsx
akurinnoy: done

PR #660, opened 2022-11-01
Sanitize workspace name

akurinnoy:
### What does this PR do?

The User Dashboard sanitizes the devworkspace name before creating the devworkspace.

fixes https://github.com/eclipse/che/issues/21750

### Is it tested? How?

- unit test added
- successfully created DW from https://github.com/kamlendu1982/tower_engin

### DIFF
--- a/packages/dashboard-frontend/src/services/helpers/__tests__/sanitizeName.spec.ts
+++ b/packages/dashboard-frontend/src/services/helpers/__tests__/sanitizeName.spec.ts
@@ -26,9 +26,9 @@ describe('sanitizeName', () => {
   });
 
   test('with non-alphanumeric character in the middle', () => {
-    const name = 'test%project';
+    const name = 'test%-#project';
     const sanitizedName = sanitizeName(name);
-    expect(sanitizedName).toEqual('test-project');
+    expect(sanitizedName).toEqual('test---project');
   });
 
   test('with non-alphanumeric first character', () => {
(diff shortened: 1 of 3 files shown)


packages/dashboard-frontend/src/services/helpers/sanitizeName.ts
```diff
@@ -0,0 +1,25 @@
...
+export default function sanitizeName(name: string): string {
+  const re = /^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$/;
+  if (re.test(name)) {
+    return name;
+  }
+
+  const sanitized = name
+    .toLowerCase()
+    .replace(/^[^a-z0-9]/, '')
+    .replace(/[^a-z0-9]$/, '')
```
olexii4, line 22: What about two or more first or last characters? It should fix but it won't. For example: 
1. ```sanitizeName(''testproject--'')``` returns ''testproject-''
2. ```sanitizeName(''--testproject'')``` returns ''-testproject''

packages/dashboard-frontend/src/services/helpers/__tests__/sanitizeName.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+  });
+
+  test('with non-alphanumeric first character', () => {
+    const name = '-test-project';
+    const sanitizedName = sanitizeName(name);
+    expect(sanitizedName).toEqual('test-project');
+  });
+
+  test('with non-alphanumeric last character', () => {
+    const name = 'test-project-';
```
olexii4, line 41: Should be checked with more than one non-alphanumeric character too. For example: 'test-project--'.

packages/dashboard-frontend/src/services/helpers/sanitizeName.ts
akurinnoy: done

olexii4 (approved):

PR #682, opened 2022-11-25
Dashboard frontend sanity check

akurinnoy:
### What does this PR do?

This PR adds the custom redux middleware to the dashboard. The middleware triggers the sanity check each time before requesting either the che-server API or the dashboard backend. The sanity check results get cached, so no unnecessary extra requests will be done.
If the sanity check fails, only one error alert will be shown. All other error alerts will be suppressed.

#### Screenshots

<details>
<summary>In preload phase</summary>

Screenshot 2022-11-28 at 17 00 31

</details>

<details>
<summary>When the dashboard loaded</summary>

Screenshot 2022-11-29 at 11 21 54
</details>

https://github.com/eclipse/che/issues/21799
	- add API sanity check with a single request to prevent showing multiple error messages
	- if after all checks above and re-tries API is still failing we show a single error with the session expired message and redirect to the login page

### Is it tested? How?

1. Open the dashboard in two windows simultaneously and sign in: https://devspaces.apps.cluster-7q8gj.7q8gj.sandbox838.opentlc.com/
2. To test this fix in preload phase: 
	- Press Ctrl(Cmd)+R in the first window and immediately sign out in the second window. 
	- You should see the error page with the corresponding message;
3. To test the fix in the running application
	- Wait at least 30 seconds after the dashboard is loaded;
	- Sign out in one of the opened windows;
	- Open the workspaces list page in the second window;
	- You should see the error alert notification.


olexii4:
I think it is not very clear to have only one temporary error message, which will be hidden after 10 seconds in the case when the session expired: 
Знімок екрана 2022-11-29 о 12 22 56
It is much better to have the message which we have when opening the 'User Preferences' page:
Знімок екрана 2022-11-29 о 12 23 48

ibuziuk:
@olexii4 @akurinnoy I actually like the current UX with `User session has expired`

packages/dashboard-frontend/src/services/bootstrap/index.ts
```diff
@@ -75,6 +71,12 @@ export default class Bootstrap {
   async init(): Promise<void> {
     this.prefetchResources();
 
+    try {
+      await this.doBackendsSanityCheck();
+    } catch (e) {
+      return;
+    }
+
```
olexii4, line 79:
```suggestion
      await this.doBackendsSanityCheck();
```

packages/dashboard-frontend/src/services/bootstrap/index.ts
```diff
@@ -250,11 +258,7 @@ export default class Bootstrap {
     try {
       await requestSettings()(this.store.dispatch, this.store.getState, undefined);
     } catch (e) {
-      this.appAlerts.showAlert({
-        key: 'bootstrap-request-workspace-settings',
-        title: common.helpers.errors.getMessage(e),
-        variant: AlertVariant.danger,
-      });
+      console.log(e);
```
olexii4, line 261:
```suggestion
      console.error(e);
```

packages/dashboard-frontend/src/services/bootstrap/index.ts
```diff
@@ -265,11 +269,7 @@ export default class Bootstrap {
     try {
       await requestServerConfig()(this.store.dispatch, this.store.getState, undefined);
     } catch (e) {
-      this.appAlerts.showAlert({
-        key: 'bootstrap-request-server-config',
-        title: common.helpers.errors.getMessage(e),
-        variant: AlertVariant.danger,
-      });
+      console.log(e);
```
olexii4, line 272:
```suggestion
      console.error(e);
```

packages/dashboard-frontend/src/services/bootstrap/index.ts
```diff
@@ -237,11 +249,7 @@ export default class Bootstrap {
     try {
       await requestNamespaces()(this.store.dispatch, this.store.getState, undefined);
     } catch (e) {
-      this.appAlerts.showAlert({
-        key: 'bootstrap-request-namespaces',
-        title: common.helpers.errors.getMessage(e),
-        variant: AlertVariant.danger,
-      });
+      console.error(e);
```
olexii4, line 252: Should we have at least one general error alert in the case when preload data is failed?

packages/dashboard-frontend/src/store/SanityCheck/index.ts
```diff
@@ -0,0 +1,138 @@
...
+        });
+
+        await axios.get('/api/kubernetes/namespace');
+        deferred.resolve(true);
+
+        dispatch({
+          type: Type.RECEIVED_BACKEND_CHECK,
+        });
+      } catch (e) {
+        const errorMessage = 'User session has expired. You need to re-login to the Dashboard.';
```
olexii4, line 95: There may be a status code different from 401. You need to check it.

akurinnoy:
@ibuziuk @olexii4 with the last fixes, the session expiration alert won't be closed by timeout.

PR #727, opened 2023-02-15
Show warning when unsupported Devfile features used

akurinnoy:
### What does this PR do?

This PR show warnings that may happen during workspace creation when the devfile uses unsupported features.

Screenshot 2023-03-03 at 18 44 54

resolves https://github.com/eclipse/che/issues/21873

### Is it tested? How?

Try creating a workspace using the following devfile: 

```yaml
schemaVersion: 2.1.0
metadata:
  generateName: unsupported-feature-
components:
  - name: web-terminal
    container:
      image: quay.io/wto/web-terminal-tooling:next
      memoryLimit: 512Mi
      mountSources: true
      command:
       - "tail"
       - "-f"
       - "/dev/null"
      dedicatedPod: true # This feature is unsupported
```

### DIFF
--- a/packages/dashboard-frontend/src/services/workspace-client/devworkspace/__tests__/devWorkspaceClient.create.spec.ts
+++ b/packages/dashboard-frontend/src/services/workspace-client/devworkspace/__tests__/devWorkspaceClient.create.spec.ts
@@ -11,9 +11,7 @@
  */
 
 import { container } from '../../../../inversify.config';
-import { DevWorkspaceBuilder } from '../../../../store/__mocks__/devWorkspaceBuilder';
 import { DevWorkspaceClient } from '../devWorkspaceClient';
-import * as DwApi from '../../../dashboard-backend-client/devWorkspaceApi';
 import * as DwtApi from '../../../dashboard-backend-client/devWorkspaceTemplateApi';
 import devfileApi from '../../../devfileApi';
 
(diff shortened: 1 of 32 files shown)


ibuziuk (approved):

ibuziuk:
@amisevsk @AObuchow folks, could you please review
I'm a bit concerned about the warning code, probably we should not expose it on the UD level

amisevsk (approved):
LGTM in general. I would ignore the HTTP status code if possible as it could be confusing to users (e.g. Kubernetes uses 299 for webhook warnings, but it's not necessarily relevant to the UI).

Worth noting: I haven't reviewed the code itself very thoroughly.

AObuchow (approved):
LGTM, though I agree with @ibuziuk and @amisevsk that I don't think the HTTP status code should be shown in the UI (if possible) as it doesn't seem very useful to the user.

packages/dashboard-frontend/src/services/workspace-client/devworkspace/__tests__/devWorkspaceClient.create.spec.ts
```diff
@@ -82,60 +89,24 @@ describe('DevWorkspace client, create', () => {
...
-        .spyOn(DwApi, 'patchWorkspace')
-        .mockResolvedValueOnce(testDevWorkspace);
-      const workspace = new DevWorkspaceBuilder()
-        .withMetadata({
-          namespace,
-          name: 'wksp-name-cvbn',
-          uid: 'uid-asdfgh12345',
-        })
-        .build();
-      spyCreateWorkspace = jest.spyOn(DwApi, 'createWorkspace').mockResolvedValueOnce(workspace);
```
olexii4, line 96: Could you remove unneeded imports('DwApi' and 'DevWorkspaceBuilder') too?

olexii4 (approved):

PR #734, opened 2023-02-22
fix: create multiarch image manifest with tools compatible with lates…

mkuznyetsov:
…t buildx extension

### What does this PR do?
'docker manifest create' won't work with images build from latest buildx extension, because of provenance attestations.
the solution is to replace this command with 'build imagetools" util that can work with images built in such way

https://github.com/eclipse/che/issues/21954

### Is it tested? How?

tested on forked repo action


nickboldt:
discussed and rejected as not necessary

PR #738, opened 2023-02-25
chore: refresh access token when workspace is starting

svor:
### What does this PR do?
Refreshes OAuth access token when the devWorkspace is starting/restarting.
It does in a few steps:
- sends a request to che-server to check the OAuth token
- if che-server returns an error, open an authentication page to authenticate

https://issues.redhat.com/browse/CRW-3928

### Is it tested? How?

1. Update CheCluster CR with `vsvydenko/che-dashboard:7` dashboard image.
2. Configure GitLab Oauth as per the documentation https://www.eclipse.org/che/docs/stable/administration-guide/configuring-oauth-2-for-gitlab/
3. Use factory to start a workspace with private project
4. Wait 2 hours when the access token is expired 
5. Try to restart the workspace and execute `git pull` in the workspace

https://user-images.githubusercontent.com/1271546/221356558-5e2e87ec-e665-4293-b4e3-72f357d80826.mp4

### DIFF
--- a/packages/dashboard-frontend/src/containers/Loader/Factory/Steps/Fetch/Devfile/index.tsx
+++ b/packages/dashboard-frontend/src/containers/Loader/Factory/Steps/Fetch/Devfile/index.tsx
@@ -227,7 +227,6 @@ class StepFetchDevfile extends AbstractLoaderStep<Props, State> {
     });
 
     try {
-      OAuthService.setOauthStartedState();
       await this.props.requestFactoryResolver(factoryUrl, params);
       this.clearNumberOfTries();
       return true;
--- a/packages/dashboard-frontend/src/services/oauth/__tests__/index.spec.ts
+++ b/packages/dashboard-frontend/src/services/oauth/__tests__/index.spec.ts
@@ -31,8 +31,24 @@ OAuthService.openOAuthPage = mockOpenOAuthPage;
 console.log = jest.fn();
 
 describe('OAuth service', () => {
-  it('should not refresh token if no projects section in devworkspace', async () => {
+  it('should not refresh token if no status section in devworkspace', async () => {
     const devWorkspace = new DevWorkspaceBuilder().build();
+
+    OAuthService.refreshTokenIfNeeded(devWorkspace);
+
+    expect(refreshFactoryOauthTokenSpy).not.toHaveBeenCalled();
+  });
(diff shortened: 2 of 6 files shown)


olexii4 (commented):
I can see some problems with refreshing the token:
```
Status Code: 500 Internal Server Error
...
Referrer Policy: strict-origin-when-cross-origin
```

Знімок екрана 2023-03-01 о 14 19 02

@akurinnoy  WDYT?

akurinnoy:
@olexii4 I also can reproduce this issue when starting workspaces with a public repo. However, this should not block this PR because the problem seems to be on the che-server side.

svor:
@olexii4 @akurinnoy thanks for reviewing 

>I can see some problems with refreshing the token:
>Status Code: 500 Internal Server Error
>...
>Referrer Policy: strict-origin-when-cross-origin

As far as I understand it's an expected behavior, we see this error because GitHub OAuth is not configured on the cluster.
This error should be caught on the dashboard side: https://github.com/eclipse-che/che-dashboard/pull/738/files#diff-c1a26f895b5713758275335441c556fffd0d06379684f741dfef6456a2866104R45-R64

@vinokurig may correct me if I'm wrong

vinokurig:
> As far as I understand it's an expected behavior, we see this error because GitHub OAuth is not configured on the cluster.
This error should be caught on the dashboard side: https://github.com/eclipse-che/che-dashboard/pull/738/files#diff-c1a26f895b5713758275335441c556fffd0d06379684f741dfef6456a2866104R45-R64
>
@vinokurig may correct me if I'm wrong

I am not sure about this error, I don't think that `che-server` can pass to dashboard errors other then `Unauthorised`.

svor:
Currently we see **Status Code: 500 Internal Server Error** when trying to restart the project without OAuth configuration:
And the exception's message is `{"message":"Unsupported OAuth provider github"}` 
Screenshot from 2023-03-06 15-05-40

With these changes a git access token is going to be refreshed each time when the workspace is starting, even if it is the first time when the workspace is starting(workspace creation) . That's why we see the exception during the workspace creation, because GitHub OAuth is not configured on the fresh OS cluster.

svor:
@akurinnoy @olexii4 @vinokurig could you please take a look

akurinnoy (approved):

PR #751, opened 2023-03-07
Fix an error when passing the editor's Devfile link

olexii4:
### What does this PR do?
Fix an error when passing the editor's Devfile link.

fixes https://github.com/eclipse/che/issues/22041


ibuziuk (approved):
:+1: would be really nice to add more tests for covering this case

PR #806, opened 2023-05-08
Configuring personal access tokens from the UD

akurinnoy:
### What does this PR do?

*todo*

resolves https://github.com/eclipse/che/issues/22112

### Is it tested? How?

*todo*


vinokurig:
I think we need a duplication check before applying a token:
1. Create a token
2. Create a new token with the same values as in the previous one.
See: Dashboard shows the list of 2 same tokens, but actually there is only one token secret in the openshift console:
screenshot-eclipse-che apps ci-ln-k69r7ft-76ef8 aws-2 ci openshift org-2023 05 17-11_26_59

akurinnoy:
@vinokurig I have just tried to follow the steps but I can't reproduce the bug. Chan you please check again that it is not a glitch?

Screenshot 2023-05-17 at 13 15 49

ibuziuk (approved):
verified and it works just fine with GitHub, also a cool feature imo is that when the token is revoked on provider side it will also disappeare from the dashboard. Azure tokens does not seem to work but it looks the problem is on che-server end and we need a separate issue for that

ibuziuk (approved):
verified and it works just fine with GitHub, also a cool feature imo is that when the token is revoked on provider side it will also disappeare from the dashboard. Azure tokens does not seem to work but it looks the problem is on che-server end and we need a separate issue for that

artaleks9:
@akurinnoy @ibuziuk 
* Verified with GitLab.com repository - works fine.
* Bitbucket.org - setup PAT is not supported in Che at all, only bitbucket server, see the doc:

-- [_Using a Git-provider access token_](https://www.eclipse.org/che/docs/stable/end-user-guide/using-a-git-provider-access-token/)

* Therefore, if this functionality plans the configuration PAT for git- servers too (bitbucket, gitlab, github), I think, need to implement/fix a several points:
-- replace Bitbucket on Bitbucker Server in the `Provider Name` drop-box
-- replace default git  provider Url on general, like this: <git_provider_url>  in the `Git Provider Endpoint`
-- if is configuring PAT for bitbucket-server should be `bitbucket-server` in the Secret, NOT `bitbucket`, because this is the wrong value and a factory won't work, see the screenshot:

che#22112-personal-access-token-bb-server


* Also, in my opinion, if a cusomer is going to add PAT for some git-server, we need to warn, that the admin should configure the ConfigMaps with self-sign certificate and CheCluster, because a factory won't work without this, see the doc:

-- [_Deploying Che with support for Git repositories with self-signed certificates_](https://www.eclipse.org/che/docs/stable/administration-guide/deploying-che-with-support-for-git-repositories-with-self-signed-certificates/)

-- _[Importing untrusted TLS certificates to Che](https://www.eclipse.org/che/docs/stable/administration-guide/importing-untrusted-tls-certificates/)_

artaleks9 (approved):

PR #837, opened 2023-06-12
[Backport] fix: apply default plugins to spec.contributions 

dkwon17:
### What does this PR do?
Backport for https://github.com/eclipse-che/che-dashboard/pull/815


ibuziuk:
@dkwon17 I believe we can close the PR and expect the fix in 3.8

PR #941, opened 2023-10-04
chore: use absolute paths for imports

akurinnoy:
### What does this PR do?

This PR forces using absolute paths for imports rather than relative ones in the `dashboard-backend` and `dashboard-frontend` packages.


olexii4 (approved):
LGTM

PR #970, opened 2023-11-02
Do not start a workspace on SCM unauthorised error

vinokurig:
### What does this PR do?
* If an unauthorised error occurs on workspace start, throw the error in order to prevent the workspace to start. 
* Do not display the authorisation error on workspace start as the page is reloaded to the SCM authorisation page.

https://issues.redhat.com/browse/CRW-4957

### Is it tested? How?

1. Deploy che from the `che-server` related PR image: `quay.io/eclipse/che-server:pr-597` see: https://github.com/eclipse-che/che-server/pull/597
2. Configure GitHub (or other) oAuth2 flow.
3. Start a workspace from the GitHub repository, accept the authorisation page.
4. Go to the GitHub settings and revoke the authorisation.
5. Remove the GitHub token in the user settings page.
6. Stop the workspace and start it again.

See: Authorisation appears and token secret is updated in the user namespace after accepting the authorization.

### DIFF
--- a/packages/dashboard-frontend/src/services/oauth/index.ts
+++ b/packages/dashboard-frontend/src/services/oauth/index.ts
@@ -63,6 +63,7 @@ export default class OAuthService {
           redirectUrl.toString(),
         );
       }
+      throw e;
     }
   }
 }
--- a/packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
+++ b/packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
@@ -365,6 +365,7 @@ export const actionCreators: ActionCreators = {
 
         getDevWorkspaceClient().checkForDevWorkspaceError(startingWorkspace);
       } catch (e) {
+        // Skip unauthorised errors. The page is redirecting to an SCM authentication page.
         if (common.helpers.errors.includesAxiosResponse(e) && isOAuthResponse(e.response.data)) {
           return;
         }
--- a/packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
+++ b/packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
@@ -32,7 +32,7 @@ import { DisposableCollection } from '@/services/helpers/disposable';
 import { FactoryParams } from '@/services/helpers/factoryFlow/buildFactoryParams';
 import { getNewerResourceVersion } from '@/services/helpers/resourceVersion';
 import { DevWorkspaceStatus } from '@/services/helpers/types';
-import OAuthService from '@/services/oauth';
+import OAuthService, { isOAuthResponse } from '@/services/oauth';
 import { loadResourcesContent } from '@/services/registry/resources';
 import { WorkspaceAdapter } from '@/services/workspace-adapter';
 import {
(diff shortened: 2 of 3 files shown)


olexii4 (approved):

packages/dashboard-frontend/src/services/oauth/index.ts
```diff
@@ -63,6 +63,7 @@ export default class OAuthService {
           redirectUrl.toString(),
         );
       }
+      throw e;
```
ibuziuk, line 66: @vinokurig @olexii4 folks, I'm a bit lost with the flow here:

- we are trying to refresh the token
- if the token is experied and we get 401, user is redirected to the OAuth auntefication page and after the redirect we throw an exception

How it used to work before, and why do we need to throw exception? Would it have some side-effects?

packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
```diff
@@ -365,6 +365,9 @@ export const actionCreators: ActionCreators = {
 
         getDevWorkspaceClient().checkForDevWorkspaceError(startingWorkspace);
       } catch (e) {
+        if (common.helpers.errors.includesAxiosResponse(e) && isOAuthResponse(e.response.data)) {
+          return;
+        }
```
ibuziuk, line 370: @vinokurig could you please add a comment in the code why we skip exception in this case?

packages/dashboard-frontend/src/services/oauth/index.ts
vinokurig: > why do we need to throw exception?

We need the exception to prevent the workspace from starting before accepting the SCM authorisation.
When dashboard opens the SCM authentication page: 
https://github.com/eclipse-che/che-dashboard/blob/c2295c4eb18d26541b27ce3b35faf8a613ec336e/packages/dashboard-frontend/src/services/oauth/index.ts#L61-L64 
the code continues to execute some time, even after the page reload command. This causes to the workspace start,  even without authorising the authentication page.
When user authorises the SCM authentication page, the SCM provider requests the callback URL and the dashboard page loads again. During the second workspace start request, we check the workspace state before requesting the refresh token API. If the workspace state is `started` the token refresh request is skipped:
https://github.com/eclipse-che/che-dashboard/blob/c2295c4eb18d26541b27ce3b35faf8a613ec336e/packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts#L291-L294
We need to throw the exception to prevent the workspace from starting, if an unauthorised error occurs.
>  How it used to work before

This bug is reproduced only with specific conditions:
* Workspace is created and started at least once.
* SCM authorisation is revoked.
* The OAuth token is removed or expired.

So looks like we have not tested this case before.
> Would it have some side-effects?

I don't think so because we just pass the error from the refresh token function further, which looks logical.

packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
vinokurig: done

ibuziuk (approved):

artaleks9:
Verified on Che with images: 
```
quay.io/eclipse/che-server:pr-597 
quay.io/eclipse/che-dashboard:pr-970
```
works as expected

artaleks9 (approved):

PR #976, opened 2023-11-06
fix: Decode spaces in factory url

tolusha:
### What does this PR do?
Decode spaces in factory url

https://github.com/eclipse/che/issues/22458

### Is it tested? How?

N/A

#### Release Notes

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,15 +98,17 @@ kubectl rollout restart deployment/che-operator -n $CHE_NAMESPACE
 
 Currently, Dashboard uses the following che-server API:
 
-| Method | Path                                |
-|--------|-------------------------------------|
-| POST    | /kubernetes/namespace/provision    |
-| GET   | /kubernetes/namespace                |
-| POST   | /factory/resolver/                  |
-| POST   | /factory/token/refresh              |
-| GET    | /oauth                              |
-| GET    | /oauth/token                        |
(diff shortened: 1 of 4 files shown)


olexii4 (approved):
LGTM

packages/dashboard-backend/src/routes/factoryAcceptanceRedirect.ts
```diff
@@ -37,6 +37,10 @@ export function registerFactoryAcceptanceRedirect(instance: FastifyInstance): vo
           params.append(ERROR_CODE_ATTR, querystring.unescape(query[ERROR_CODE_ATTR] as string));
         }
 
+        const url = params.get('url');
+        if (url) {
+          params.set('url', url.replaceAll(' ', '%20'));
```
ibuziuk, line 42: can we use `encodeURIComponent` to cover the general case? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent

akurinnoy (approved):
LGTM

ibuziuk (approved):
I'm ok to merge as a hot fix for 3.10, but not really understand why do we decoding twice in general🤷

akurinnoy (approved):

PR #987, opened 2023-11-14
Do not throw other then unauthorised error on token refresh

vinokurig:
### What does this PR do?
Do not throw exceptions other then unauthorised exception on token refresh request when starting a workspace.

https://issues.redhat.com/browse/CRW-5015

### Is it tested? How?

1. Make sure GitHub oauth is **NOT** configured
2. Start a workspace from a public GitHub repository.
3. Stop the workspace and start it again.

See: workspace starts without any issues.

### DIFF
--- a/packages/dashboard-frontend/src/services/oauth/index.ts
+++ b/packages/dashboard-frontend/src/services/oauth/index.ts
@@ -62,8 +62,8 @@ export default class OAuthService {
           response.data.attributes.oauth_authentication_url,
           redirectUrl.toString(),
         );
+        throw e;
       }
-      throw e;
     }
   }
 }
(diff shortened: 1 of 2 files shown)


artaleks9:
Verified on Che with `quay.io/eclipse/che-dashboard:pr-987` image.
The functionality works as expected.

packages/dashboard-frontend/src/services/oauth/index.ts
```diff
@@ -62,8 +62,8 @@ export default class OAuthService {
           response.data.attributes.oauth_authentication_url,
           redirectUrl.toString(),
         );
+        throw e;
       }
-      throw e;
```
ibuziuk, line 66: @vinokurig would it be possible to add test for this functionality? Do we need to re-throw exception in general?

vinokurig: > would it be possible to add test for this functionality?

done
>Do we need to re-throw exception in general?

Yes, the exception interrupts the workspace start event until the authorisation callback.

akurinnoy (approved):

tolusha (approved):

ibuziuk (approved):

PR #992, opened 2023-11-16
Update yarn.lock file to get 7.77.0 version of che-devworkspace-gener…

akurinnoy:
### What does this PR do?

This is to fix dependent DevSpaces builds which are failing because the offline build predownloads 7.77.0-next-363e98c then tries to build with 7.77.0.


ibuziuk (approved):

nickboldt (approved):
seems OK

PR #995, opened 2023-11-17
Cleanups

akurinnoy:
### What does this PR do?

- get rid of the PR check running on Node.JS 16;
- cleanup dependencies.


packages/dashboard-backend/src/routes/api/devworkspaceResources.ts
```diff
@@ -42,7 +42,11 @@ export function registerDevworkspaceResourcesRoute(instance: FastifyInstance) {
             editorContent,
             projects: [],
           },
-          axiosInstance,
+          // because of the axios version mismatch between the generator and the dashboard
+          // we need to cast the axiosInstance to any
+          // the difference is the `method` property type which is `Method | undefined` in the generator
+          // and `Method | string | undefined` in the dashboard
+          axiosInstance as any,
```
olexii4, line 49: We definitely need to upgrade https://github.com/eclipse-che/che-devfile-registry/blob/main/tools/devworkspace-generator/package.json#L45 instead of casting to any.
WDYT?

```
"axios": "1.6.0",
```
-> 
```
"axios": "^1.6.0",
```
or
```
"axios": "1.6.2",
```

PR #1016, opened 2023-12-06
Add attribute controller.devfile.io/bootstrap-devworkspace: true when Devfile resolution via API fails or is not supported

akurinnoy:
### What does this PR do?

This PR adds the attribute `controller.devfile.io/bootstrap-devworkspace: true` when Devfile resolution via API fails or is not supported. 

resolves https://github.com/eclipse/che/issues/22697

### Is it tested? How?

1. Deploy Eclipse Che and the dashboard image from this PR:
  ```sh
  kubectl patch -n eclipse-che "checluster/eclipse-che" --type=json -p="[{"op": "replace", "path": "/spec/components/dashboard/deployment", "value": {containers: [{image: "quay.io/eclipse/che-dashboard:pr-1016", name: che-dashboard}]}}]"
  ```
2. Add your SSH keys to the Dashboard, and make sure the public one is uploaded to your Git provider.
3. Copy the git+SSH URL of your git repository, and create a new workspace using this URL.
4. The workspace should start successfully with the project cloned and the Devfile applied.


olexii4 (approved):
LGTM

PR #1063, opened 2024-03-06
Fix support for URL params inside the URL of getting started samples definition

olexii4:
### What does this PR do?
Fix support for URL params inside the URL of getting started samples definition.

fixes https://github.com/eclipse/che/issues/22866

### DIFF
--- a/packages/dashboard-frontend/src/pages/GetStarted/SamplesList/__tests__/index.spec.tsx
+++ b/packages/dashboard-frontend/src/pages/GetStarted/SamplesList/__tests__/index.spec.tsx
@@ -51,7 +51,7 @@ describe('Samples List', () => {
                 tags: ['Community', 'Java', 'Quarkus', 'OpenJDK', 'Maven', 'Debian'],
                 icon: '/images/quarkus.svg',
                 links: {
-                  v2: sampleUrl,
+                  v2: `${sampleUrl}?df=devfile2.yaml'`,
                   devWorkspaces: {
                     'che-incubator/che-code/insiders':
                       'registry-url/devfile-registry/devfiles/quarkus/devworkspace-che-code-insiders.yaml',
@@ -106,7 +106,7 @@ describe('Samples List', () => {
       userEvent.click(sampleCardButton);
 
       expect(mockWindowOpen).toHaveBeenCalledWith(
-        `/load-factory?url=${encodeURIComponent(
+        `/load-factory?df=devfile2.yaml%27&url=${encodeURIComponent(
           sampleUrl,
         )}&che-editor=che-incubator%2Fche-code%2Finsiders&devWorkspace=registry-url%2Fdevfile-registry%2Fdevfiles%2Fquarkus%2Fdevworkspace-che-code-insiders.yaml&editor-image=custom-editor-image&storageType=${preferredPvcStrategy}`,
         '_blank',
@@ -126,7 +126,7 @@ describe('Samples List', () => {
       userEvent.click(sampleCardButton);
 
       expect(mockWindowOpen).toHaveBeenCalledWith(
-        `/load-factory?url=${encodeURIComponent(
+        `/load-factory?df=devfile2.yaml%27&url=${encodeURIComponent(
           sampleUrl,
         )}&che-editor=che-incubator%2Fche-code%2Finsiders&devWorkspace=registry-url%2Fdevfile-registry%2Fdevfiles%2Fquarkus%2Fdevworkspace-che-code-insiders.yaml&editor-image=custom-editor-image&storageType=ephemeral`,
         '_blank',
@@ -169,7 +169,7 @@ describe('Samples List', () => {
       userEvent.click(sampleCardButton);
 
       expect(mockWindowOpen).toHaveBeenCalledWith(
-        `/load-factory?url=${encodeURIComponent(
+        `/load-factory?df=devfile2.yaml%27&url=${encodeURIComponent(
           sampleUrl,
         )}&che-editor=che-incubator%2Fche-code%2Finsiders&devWorkspace=registry-url%2Fdevfile-registry%2Fdevfiles%2Fquarkus%2Fdevworkspace-che-code-insiders.yaml&editor-image=custom-editor-image&storageType=${preferredPvcStrategy}`,
         '_blank',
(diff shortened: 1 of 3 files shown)


packages/dashboard-frontend/src/pages/GetStarted/SamplesList/index.tsx
```diff
@@ -25,7 +25,7 @@ import { connect, ConnectedProps } from 'react-redux';
 
 import SamplesListGallery from '@/pages/GetStarted/SamplesList/Gallery';
 import SamplesListToolbar from '@/pages/GetStarted/SamplesList/Toolbar';
-import { EDITOR_ATTR, EDITOR_IMAGE_ATTR } from '@/services/helpers/factoryFlow/buildFactoryParams';
+import { EDITOR_ATTR, EDITOR_IMAGE_ATTR, FACTORY_URL_ATTR } from "@/services/helpers/factoryFlow/buildFactoryParams";
```
akurinnoy: Please run `yarn lint:fix`

packages/dashboard-frontend/src/pages/GetStarted/SamplesList/index.tsx
```diff
@@ -72,7 +72,12 @@ class SamplesList extends React.PureComponent<Props, State> {
   private async handleSampleCardClick(metadata: DevfileRegistryMetadata): Promise<void> {
     const { editorDefinition, editorImage } = this.props;
 
-    const factoryUrlParams = new URLSearchParams({ url: metadata.links.v2 });
+    const url = new URL(metadata.links.v2);
+
+    const factoryUrlParams = new URLSearchParams(url.searchParams);
+
+    url.search = '';
```
akurinnoy, line 83: Are you sure it's safe to drop all of the search parameters? I think some of them may be necessary for proper factory URL resolution.
My suggestion is to remove only parameters known by the dashboard (e.g. `df`, `che-editor`, `editor-image`) and keep the rest as part of the factory URL.

packages/dashboard-frontend/src/pages/GetStarted/SamplesList/__tests__/index.spec.tsx
```diff
@@ -51,7 +51,7 @@ describe('Samples List', () => {
                 tags: ['Community', 'Java', 'Quarkus', 'OpenJDK', 'Maven', 'Debian'],
                 icon: '/images/quarkus.svg',
                 links: {
-                  v2: sampleUrl,
+                  v2: `${sampleUrl}?df=devfile2.yaml'`,
```
akurinnoy, line 54:
```suggestion
                  v2: `${sampleUrl}?df=devfile2.yaml`,
```

packages/dashboard-frontend/src/pages/GetStarted/SamplesList/index.tsx
olexii4: I am sure. I implemented the same behavior that we have for a factory:
```{che-server}/dashboard/load-factory?url=```

I can explain. All requests will be sent to ```{che-server}/api/factory/resolver ```, but che-server supported only several parameters like this one ```override.metadata.generateName```.

Existing implementation propagates allowed parameters to the factory resolver.

For example:
```
{che-server}/dashboard/load-factory?url=https://github.com/eclipse-che/che-dashboard&override.metadata.generateName=new-name
```
Знімок екрана 2024-03-06 о 21 01 26

Try it.

Comment (akurinnoy) [reply packages/dashboard-frontend/src/pages/GetStarted/SamplesList/index.tsx:83]: I think this example is not relevant to the issue you're fixing. Different components are involved depending on the workspace creation way: sample card or git URL. Have you tried to create a workspace from a sample with the same URL?

**Comment by akurinnoy** [comment]
@olexii4, 
I tested this PR and it doesn't seem to fully fix the original issue.

The issue description states that the sample with this URL `https://example.com/my-project.git?df=next/base/devfile.yaml` didn’t work correctly.

To test this case I prepared the following sample file JSON:

```json
[
	{
		"displayName": "Test Sample",
		"description": "The sample to test URL with query parameters",
		"tags": ["Empty"],
		"url": "https://github.com/akurinnoy/vigilant-umbrella?df=other.devfile.yaml"
	}
]
```

And then I proceeded steps 3-6 from How to test section in this PR.
I expected a new workspace to be created with the name `vigilant-umbrella-other-xxxx` according to the `generateName` field in the `other.devfile.yaml`. But the devfile was not resolved correctly. Instead, the default devfile was used.

packages/dashboard-frontend/src/pages/GetStarted/SamplesList/index.tsx
olexii4: I think this example is relevant to your suggestion to remove only parameters the dashboard knows (e.g. df, che-editor, editor-image) and keep the rest as part of the factory URL.

olexii4:
> @olexii4, I tested this PR and it doesn't seem to fully fix the original issue.
> 
> The issue description states that the sample with this URL `https://example.com/my-project.git?df=next/base/devfile.yaml` didn’t work correctly.
> 
> To test this case I prepared the following sample file JSON:
> 
> ```json
> [
> 	{
> 		"displayName": "Test Sample",
> 		"description": "The sample to test URL with query parameters",
> 		"tags": ["Empty"],
> 		"url": "https://github.com/akurinnoy/vigilant-umbrella?df=other.devfile.yaml"
> 	}
> ]
> ```
> 
> And then I proceeded steps 3-6 from How to test section in this PR. I expected a new workspace to be created with the name `vigilant-umbrella-other-xxxx` according to the `generateName` field in the `other.devfile.yaml`. But the devfile was not resolved correctly. Instead, the default devfile was used.

@akurinnoy For the first implementation it expected 
```https://github.com/akurinnoy/vigilant-umbrella?override.devfileFilename=other.devfile.yaml```
 instead of 
```https://github.com/akurinnoy/vigilant-umbrella?df=other.devfile.yaml```

I have fixed it.

akurinnoy:
@olexii4 Thanks!

packages/dashboard-frontend/src/pages/GetStarted/SamplesList/index.tsx
akurinnoy: got it, thanks

akurinnoy (approved):

PR #1084, opened 2024-04-03
Fix the blank page if DevWorkspace is missing controller.devfile.io/creator label

akurinnoy:
### What does this PR do?

With this fix the dashboard can correctly process DevWorkspaces without `metadata.label`.

fixes https://github.com/eclipse/che/issues/22899

### Is it tested? How?

See "Steps to reproduce" [in the issue](https://github.com/eclipse/che/issues/22899).


olexii4 (approved):
LGTM

akurinnoy:
@ibuziuk That was an assumption that all devworkspaces should have a label field.

ibuziuk (approved):

PR #1088, opened 2024-04-04
Changes Podman's priority for the container_tools script

olexii4:
### What does this PR do?
This PR changes Podman's priority for the container_tools script.  It allows users to use Podman as default if both tools are installed and running.


akurinnoy (approved):

ibuziuk (commented):
looks like PR check is failing

PR #1119, opened 2024-05-29
Remove the deprecation warning message in the dashboard pod

akurinnoy:
### What does this PR do?

This PR changes the dashboard backend build so the resulting js bundle has no dependency packages included.

### Is it tested? How?

1. Deploy Eclipse Che and patch CR with this command: **todo**
2. Create and launch a workspace.
3. Open the dashboard pod logs, there should be no deprecation warnings.


olexii4 (approved):
LGTM

PR #1141, opened 2024-07-03
fix: If persistHome is enabled, the token in .kube/config isn't renewed

batleforc:
### What does this PR do?

This PR has for goal to fix [this issue](https://github.com/eclipse-che/che/issues/22924).
The aforementioned issue present a situation where if the persist Home option is enabled in che's conf, the kubeconfig that should be in ~/.kube/config isn't updated any more with an up-to-date token. This situation hinder further work and prevent the monitoring extension in VsCode to work properly.

So, this PR patch the function that should inject into each container the Kubeconfig with:
- A check that will not inject the kubeconfig in the case where the user manually inject the kubeconfig 
- A fetch that will take the existing kubeconfig inside the container and try to merge it with the generated one. The merge is based on the name, if a block with the same name exist in the existing kubeconfig it will be replaced. 

<https://github.com/eclipse-che/che/issues/22924>

### Is it tested? How?

The Lerna  (Test/Lint) command has been executed, and I set it up in my own che server.

#### Release Notes

N/A

#### Docs PR

I don't think that this should add a doc entry has it is a bug.

### DIFF
--- a/packages/dashboard-backend/src/devworkspaceClient/services/kubeConfigApi.ts
+++ b/packages/dashboard-backend/src/devworkspaceClient/services/kubeConfigApi.ts
@@ -76,6 +76,11 @@ export class KubeConfigApiService implements IKubeConfigApi {
         );
 
         // if -f ${kubeConfigDirectory}/config is not found then sync kubeconfig to the container
+        // https://github.com/eclipse-che/che/issues/22924#issuecomment-2059944557
+        // TODO:
+        //  - Create a function that will check if the kubeconfig look's like one created by this function
+        //  - Check if the kubeconfig is mounted from a secret/configmap
+        //  - Check if the token is valid, if not replace the config
         const kubeConfig = this.setNamespaceInContext(this.kubeConfig, namespace);
         await exec(
           podName,
--- a/packages/dashboard-backend/src/devworkspaceClient/services/kubeConfigApi.ts
+++ b/packages/dashboard-backend/src/devworkspaceClient/services/kubeConfigApi.ts
@@ -75,13 +75,31 @@ export class KubeConfigApiService implements IKubeConfigApi {
           this.getServerConfig(),
         );
 
-        // if -f ${kubeConfigDirectory}/config is not found then sync kubeconfig to the container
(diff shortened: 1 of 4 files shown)


akurinnoy:
Hi @batleforc, thanks for contributing!
Please sign the Eclipse Contributor Agreement [here](https://accounts.eclipse.org/user/5930/eca/3.1.0) and sign off your commits, as this is required to unlock the PR merge.

batleforc:
Hi @akurinnoy
It should be good, thanks for the quick answer

packages/dashboard-backend/src/devworkspaceClient/services/kubeConfigApi.ts
```diff
@@ -75,8 +75,31 @@ export class KubeConfigApiService implements IKubeConfigApi {
           this.getServerConfig(),
         );
 
-        // if -f ${kubeConfigDirectory}/config is not found then sync kubeconfig to the container
-        const kubeConfig = this.setNamespaceInContext(this.kubeConfig, namespace);
+        // check if the kubeconfig is already mounted
+        if (container.volumeMounts?.some(vm => vm.mountPath.includes(kubeConfigDirectory))) {
```
tolusha, line 79: ?
```suggestion
        if (container.volumeMounts?.some(vm => vm.mountPath === kubeConfigDirectory)) {
```

tolusha:
Sounds good for me.
I think it is better to add some tests as well

packages/dashboard-backend/src/devworkspaceClient/services/kubeConfigApi.ts
batleforc: I just committed it. At first, I thought that the include would be better in case of deep path kubeconfig but didn't think about the possible case where the path without any link could include this kind of path

akurinnoy:
Hi @batleforc , there is a failed unit test in the dashboard-backend package, could you look at it?

batleforc:
Test has been fixed and i wrote the corresponding test in the .spec part

akurinnoy (approved):

olexii4 (approved):

PR #1145, opened 2024-07-09
fix: stop using mirror.centos.org repo

artaleks9:
### What does this PR do?
* Replaces usage of http://mirrorlist.centos.org/ repo with http://vault.centos.org/ as http://mirrorlist.centos.org/ is now down that Centos 7 is EOL as of July 1st 2024.
* Without this fix, the Dockerfile fails to build on [Openshift CI](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/eclipse-che_che-dashboard/1144/pull-ci-eclipse-che-che-dashboard-main-v14-dashboard-happy-path/1809377858603192320) with the error:
```
Could not retrieve mirrorlist http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=os&infra=container error was
14: curl#6 - "Could not resolve host: mirrorlist.centos.org; Unknown error" 
```
* This is more of a temporary work-around as the issue probably should be fixed in the image being used:
`registry.ci.openshift.org/openshift/release:golang-1.20`

N/A

### Is it tested? How?

N/A

### DIFF
--- a/.ci/openshift-ci/Dockerfile
+++ b/.ci/openshift-ci/Dockerfile
@@ -11,12 +11,13 @@
 # https://github.com/eclipse-che/che-release/blob/main/.github/workflows/update-base-images.yml
 FROM registry.ci.openshift.org/openshift/release:golang-1.20
 
+SHELL ["/bin/bash", "-c"]
+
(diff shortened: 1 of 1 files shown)


akurinnoy (approved):
LGTM

PR #1179, opened 2024-08-30
Show a warning notification on fetch token error when starting a work…

vinokurig:
…space

### What does this PR do?
Show a warning notification on fetch token error when starting a workspace.

https://github.com/eclipse-che/che/issues/23076

### Is it tested? How?

see https://github.com/eclipse-che/che-server/pull/715

### DIFF
--- a/packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
+++ b/packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
@@ -12,6 +12,7 @@
 
 import { V1alpha2DevWorkspaceStatus } from '@devfile/api';
 import common, { api, ApplicationId } from '@eclipse-che/common';
+import { includesAxiosResponse } from '@eclipse-che/common/lib/helpers/errors';
 import { dump } from 'js-yaml';
 import { Action, Reducer } from 'redux';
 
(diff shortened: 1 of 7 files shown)


ibuziuk (approved):

packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
```diff
@@ -311,7 +311,32 @@ export const actionCreators: ActionCreators = {
         return;
       }
       try {
-        await OAuthService.refreshTokenIfProjectExists(workspace);
+        try {
```
olexii4, line 314: @vinokurig Could we move it out from the previous try-catch block?

Could you add proper tests like this one https://github.com/eclipse-che/che-dashboard/blob/main/packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/__tests__/actions.spec.ts#L405?

ibuziuk:
@olexii4 @akurinnoy since Igor is on PTO this week, we should either merge it as is if in principal all good, or somebody from dashboard side need to provide update on this PR with required changes

packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
vinokurig: > Could we move it out from the previous try-catch block?

No we can not, otherwise the workspace will not proceed to start if the `refreshTokenIfProjectExists` function fails.

vinokurig: Sorry for confusion, fixed.

packages/dashboard-frontend/src/store/Workspaces/devWorkspaces/index.ts
```diff
@@ -311,7 +311,32 @@ export const actionCreators: ActionCreators = {
...
+          await OAuthService.refreshTokenIfProjectExists(workspace);
+        } catch (e: any) {
```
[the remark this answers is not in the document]
vinokurig: done

olexii4 (approved):

PR #1207, opened 2024-09-30
fix: Unexpected error message No IDE URL after 20 sec during workspace startup

olexii4:
### What does this PR do?
These changes fixed problems with a workspace status which may flicker from starting to running and back to starting.
A special rule was added and a timer will be restarted each time when the workspace status flicks from starting to running.

Знімок екрана 2024-09-30 о 16 50 32

fixes https://github.com/eclipse-che/che/issues/23159

### Is it tested? How?

See **Steps to reproduce**  from https://github.com/eclipse-che/che/issues/23159


ibuziuk (approved):
please, backport to 7.92.x as well

akurinnoy (approved):

PR #1249, opened 2024-11-05
Add 'experimental' placeholder to the passphrase

ibuziuk:
### What does this PR do?

See https://github.com/devfile/devworkspace-operator/issues/1337#issuecomment-2457461589

Related to https://github.com/devfile/devworkspace-operator/issues/1337


dmytro-ndp (approved):
Good idea, IMHO

olexii4 (approved):

vinokurig (changes requested):
From the [discussion](https://github.com/devfile/devworkspace-operator/issues/1337#issuecomment-2457442532) we agreed only to [mark the injection of the SSH agent postStart event as experimental](https://github.com/devfile/devworkspace-operator/pull/1344). The SSH agent post start event is needed only for the SSH flow from terminal, so disabling the particular feature will not affect the rest ssh features e.g. start workspace from ssh url or [git remote operations from che-code](https://github.com/che-incubator/che-code/pull/400). So I would rather add some documentation regarding the SSH flow from terminal e.g. we need to enable DWO experimental features to have this feature.

PR #1313, opened 2025-02-10
Change the git-userdata configmap to secret

vinokurig:
### What does this PR do?
Store the gitconfig automaunt content to secret instead of configmap. This is needed to have all gitconfig automaunt values in one secret as devworkspace operator does not support mount to gitconfig from multiple objects.

The goal of having a single place to mount gitconfig values is that we need to mount additional gitconfig section apart from the `[user]` section. Currently we mount only username and email from the `workspace-userdata-gitconfig-configmap` configmap but in the scope of the[ Azure devops server support](https://github.com/eclipse-che/che-server/pull/754) we also need to mount azure token as an http.extraHeader value:
```
[http]
    extraHeader: Bearer <azure token>
``` 
We can not mount the `http` section in a separate secret because devworkspace operator does not support gitconfig mount from multiple objects, see https://github.com/devfile/devworkspace-operator/blob/a6ec0bfb254ae1f63283b507475fba69b9768ac5/pkg/provision/automount/common.go#L223-L225.
https://github.com/eclipse-che/che/issues/23306

### Is it tested? How?

1. Go to the user preferences -> Gitconfig
2. Set up username and email

See: the data is stored to the `devworkspace-gitconfig-automaunt-secret` Secret in the user namespace.

### DIFF
--- a/packages/dashboard-backend/src/devworkspaceClient/services/gitConfigApi/index.ts
+++ b/packages/dashboard-backend/src/devworkspaceClient/services/gitConfigApi/index.ts
@@ -21,7 +21,7 @@ import {
 } from '@/devworkspaceClient/services/helpers/prepareCoreV1API';
 import { IGitConfigApi } from '@/devworkspaceClient/types';
 
-const GITCONFIG_CONFIGMAP = 'workspace-userdata-gitconfig-configmap';
+const GITCONFIG_SECRET = 'devworkspace-gitconfig-automaunt-secret';
 const GITCONFIG_API_ERROR_LABEL = 'CORE_V1_API_ERROR';
 
 export class GitConfigApiService implements IGitConfigApi {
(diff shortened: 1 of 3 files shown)


ibuziuk (commented):
@vinokurig what about backward compatibility?

vinokurig:
@ibuziuk 
> what about backward compatibility?

We will not have the backward compatibility in this case. To have it we can:
* Use the current `workspace-userdata-gitconfig-configmap` as the single place to store the gitconfig mount. cons: 
  * Sensitiva data such as tokens should be stored in a secret.
  * The configmap name is too specific to store other then username and email data.
* Deprecate the configmap and add a temporary mechanism that will transfer data from the configmap to the new secret.
WDYT?

olexii4 (commented):
@vinokurig  I tested this PR with Dashboard Version 7.99.0-next.

Before update:
Знімок екрана 2025-02-10 о 21 22 36

After update:
Знімок екрана 2025-02-10 о 21 22 43

Let's read the **git-userdata**(if they exist) from **configmap** and store it into **secret**. Then, removed the **git-userdata** from **configmap**.

WDYT?

vinokurig:
@olexii4 I have fixed it in the related che-server pull request https://github.com/eclipse-che/che-server/pull/754

olexii4 (approved):
LGTM

tolusha:
I faced the same problem, I can't see gitconfig after switching to the new image.
BTW, we have to update the doc as well
https://eclipse.dev/che/docs/stable/end-user-guide/mounting-git-configuration/

packages/dashboard-backend/src/devworkspaceClient/services/gitConfigApi/index.ts
```diff
@@ -37,13 +37,13 @@ export class GitConfigApiService implements IGitConfigApi {
...
+      const response = await this.coreV1API.readNamespacedSecret(GITCONFIG_SECRET, namespace);
 
       return this.toGitConfig(response.body);
     } catch (error) {
       if (helpers.errors.isKubeClientError(error) && error.statusCode === 404) {
-        // Create gitconfig configmap if it does not exist
-        return this.createGitConfigMap(namespace);
+        // Create gitconfig secret if it does not exist
+        return this.createGitConfigSecret(namespace);
       }
```
ibuziuk, line 47: I'm wondering how this change would be working from the backward compatibility perspective?
e.g. if there is a cm with gitconfig once the new version of the project is rolled out would the gitconfig be blank?

PR #1357, opened 2025-06-23
Hiding the JetBrains editor by default on P/Z

prabhuk25:


packages/dashboard-backend/src/routes/api/clusterConfig.ts
```diff
@@ -28,6 +33,64 @@ export function registerClusterConfigRoute(instance: FastifyInstance) {
   });
 }
 
+async function getArchitectureFromDashboardPod(): Promise<string> {
```
olexii4, line 36: @prabhuk25 Thanks for your contribution.

I have some feedback on this code:

1. This is not the correct place for such code. You can review how `getAllWorkspacesLimit` and `getDashboardLogo` were implemented.

2. You can read the current infrastructure from the che-dashboard pod by simply running the **arch** command line command with Node.js.

3. I think a simple name, `getCurrentArchitecture`, would be better.

olexii4: I have created a small [PR](https://github.com/eclipse-che/che-dashboard/pull/1360) as an example of how to extend the dashboard-backend API.

PR #1359, opened 2025-06-24
Better handling of the che-workspaces link 

olexii4:
### What does this PR do?
Better handling of the che-workspaces link 

fixes https://github.com/eclipse-che/che/issues/23402


svor (commented):
@olexii4 
I tried with `new` parameter - works good
When use `https://github.com/eclipse-che/che-docs?existing=che-doc` - it opens an existed workspace as expected

But when i try to skip workspace name or use not existed workspace name - workspace is not started. Examples:
```
https://eclipse-che-instance/dashboard/#https://github.com/eclipse-che/che-docs?existing
https://github.com/eclipse-che/che-docs?existing=name-doesnt-exist
``` 
but as I understand, the first workspace with the same source link should be started. I just see this screen:
screenshot-mail_google_com-2025_07_03-12_44_19

vinokurig:
The functionality does not work properly with repositories without devfie:
1. Create first workspace from a repository without devfile.
2. Create second workspace from the same repository:
image

vinokurig:
I believe this approach with URL query parameters is not obvious to a regular user. This approach removes the interactive way to start a new workspace (the warning popup). So, if a user creates a new workspace with the same name, they will not have the option to create a new one unless they know the URL query parameters. @ibuziuk WDYT?

olexii4:
@svor I have added a UX mockup for the flow: choice of which workspace to pick, if multiple workspaces for that project have been created.

Знімок екрана 2025-07-28 о 06 13 04

olexii4:
> @olexii4 Nice improvements, I have a few comments: In the view, where the user can select workspace to start, we should use Warning icon instead of Error icon together with `Checking if a workspace with the same name exists`: screenshot-che-dogfooding_apps_che-dev_x6e0_p1_openshiftapps_com-2025_07_28-12_31_39
> 
> Also it's important to add an information about a new switch button on the Create Workspace page to the description. screenshot-che-dogfooding_apps_che-dev_x6e0_p1_openshiftapps_com-2025_07_28-12_53_22 About this switcher. As for me it should have ON state by default to avoid braking current flow of creating workspaces from the sample. The name can be `Create New` and we can add **?** icon with the tooltip: `Create a new workspace each time when ON; reuse an existing one when OFF.`

@svor, I have fixed the `Create New` switcher.

The current implementation of the progress tree supports just two types of icons:

Знімок екрана 2025-07-29 о 16 26 49

 I think it would be better to create a new issue for adding support for the new type of icon for the progress tree and fix it separately from this PR. 

WDYT?

svor:
> I think it would be better to create a new issue for adding support for the new type of icon for the progress tree and fix it separately from this PR.

@olexii4 please create a separate issue

vinokurig:
Tested the functionality; it works as described in the description. However, my concern is that to use the widget, a user has to start a workspace with the `?new` query parameter, which might not be obvious to a regular user.
I prefer to keep the old way of notifying, but to add the ability to choose from the list of existing workspaces. With this approach, a user will be able to create new workspaces without the query parameter.

vinokurig (approved):

dmytro-ndp:
@olexii4: thank you for the improvement of the User Dashboard UX!

Here is the PR validation summary:
1) test scenarios described in the description of this PR have passed with Eclipse Che Next using quay.io/eclipse/che-dashboard:pr-1359-amd64 with Empty workspace and sample workspaces
2) target issue https://github.com/eclipse-che/che/issues/23402 has been mostly resolved, apart from a few cases described below.

### Test case when only one workspace can be running at a time:
1. Install Eclipse Che Next using quay.io/eclipse/che-dashboard:pr-1359-amd64 with default settings, where only one workspace could be running at a time.
2. Go to User Dashboard and ensure "Create New" check box is disabled by default.
3. Create a workspace from **.Net 5.0** sample.
4. Go to the workspace sample list and click on **.Net 5.0** sample tile again.

**Expected behavior**:
User Dashboard redirects the user to the existing and running **.Net 5.0** workspace.

**Actual behavior**:
Start workspace page showed a warning message:
> Running workspace(s) found.
> You can only have 1 running workspace at a time.
Screenshot from 2025-08-05 22-33-51

### Test case for creating multiple workspaces from the same Git devfile URL with "Create New" enabled
1. Install Eclipse Che Next using quay.io/eclipse/che-dashboard:pr-1359-amd64.
2. Go to the User Dashboard and enable the "Create New" checkbox.
3. Create a workspace from the following URL: https://github.com/eclipse-che/che-dashboard/blob/main/devfile.yaml.
4. Stop the created workspace and create a second one from the same URL.
5. Stop the second workspace and create a third one from the same URL.

**Expected Behavior**:
The User Dashboard creates a second and third workspaces from the URL https://github.com/eclipse-che/che-dashboard/blob/main/devfile.yaml without any warnings.

**Actual Behavior**:
- **At step 4**, the "Start Workspace" page showed a warning when creating the new workspace:
> Existing workspace found
> A workspace with the same name (che-dashboard) has been found. Should you want to open the existing workspace or proceed to create a new one, please choose the corresponding action.
Screenshot from 2025-08-06 02-20-46

- **At step 5**, the User Dashboard did not create a third workspace; instead, it started the second one without any warnings.
_Note_: The same behavior was observed at step 5 when the "**Create New**" checkbox was disabled.

svor:
@dmytro-ndp good catch! +1 to remove **Create New If Existing** from the **Import from Git** menu

dmytro-ndp (approved):

PR #1388, opened 2025-09-30
fix: Remove controller.devfile.io/scc attribute only if exists

tolusha:
### What does this PR do?
Remove controller.devfile.io/scc attribute only if exists

https://github.com/eclipse-che/che/issues/23592


olexii4 (approved):
LGTM

PR #1505, opened 2026-03-30
feat: AI Selector Widget

### LINKED ISSUE #23796: [UD] Add AI coding tool selection and API key management for DevWorkspaces
### Is your enhancement related to a problem? Please describe

Eclipse Che Dashboard has no built-in way for users to select an AI coding tool or manage API keys for AI providers. Users must manually configure DevWorkspace specs and create Kubernetes Secrets, which is error-prone and not discoverable.

#### 1. AI provider and tool definitions

Separate provider metadata (name, icon, description) from tool injection details (container image, binary, env var). Multiple tools can share one provider. Definitions are served to the frontend via the server config endpoint and can be overridden by administrators through the CheCluster CR.

#### 2. AI provider selector on workspace creation

Add a selector on the Get Started page so users can pick an AI provider when creating a workspace. The selected provider ID is passed as a factory URL parameter and the tool is injected into the DevWorkspace at creation time.

#### 3. API key management in User Preferences

Add an "AI Providers Keys" tab in User Preferences for managing API keys:

- List configured keys with provider name and status
- Add, update, or delete keys

Keys are stored as Kubernetes Secrets with auto-mount labels so the DevWorkspace Controller injects them as environment variables.

#### 4. AI tool info on Workspace Details page

Show the currently injected AI tool in the Overview tab. Allow users to view available tools and switch to a different one (workspace must be stopped).

#### 5. Backend API

- Serve provider and tool definitions via server config
- REST endpoints for listing, creating, and deleting API key secrets scoped to the user's namespace

### Describe alternatives you've considered

- **CheCluster CR only:** Administrators configure AI tools entirely via CR, but users get no agency to select tools or manage their own API keys.
- **Flat tool definitions:** Keeping provider metadata on the tool definition duplicates data when multiple tools share a provider.

olexii4:
### What does this PR do?
Adds an **AI Selector widget** to the Create Workspace page and integrates AI tool management across the dashboard, allowing users to pick an AI tool (e.g. Gemini CLI, Claude Code, OpenCode), save their API key, and have the tool automatically injected into workspaces.

#### Type Definitions

Two separate types model the AI feature:

- **`AiProviderDefinition`** — provider metadata (id, name, publisher, description, icon, docs URL). Icons and descriptions live on the provider, not on individual tools.
- **`AiToolDefinition`** — injection details (providerId as FK to provider, tag, binary, pattern, injectorImage, envVarName). The `providerId` follows `vendor/product` convention (e.g. `anthropic/claude`, `google/gemini`).

#### Create Workspace — AI Provider Selector
- Accordion-based UI ("Use a Default AI Provider" / "Choose an AI Provider") rendered on the Create Workspace page between Editor Selector and Import form.
- Link to "Manage API keys in User Preferences" for quick navigation.

#### User Preferences — AI Providers Keys
- New **"AI Providers Keys"** tab in User Preferences with a table listing configured API keys per tool.
- Add/Edit modal with provider dropdown and masked API key input; 
- Delete confirmation modal.

#### Workspace Details — Overview Tab
- `AiToolFormGroup` section displaying the currently injected AI tool (or "None").
- Edit mode opens a radio-button modal listing all available tools with **descriptions from the provider definition**.
- Info modal explaining the injection mechanism and listing available tools with provider descriptions.

#### Workspaces List — AI Provider Column
- New "AI Provider" column in the Workspaces List table.

#### AI Tool Injection into DevWorkspace
- `addAiToolToWorkspace()` creates an init container from the tool's `injectorImage`, a shared `injected-tools` volume, and DevWorkspace lifecycle commands:
  - **preStart**: `apply` command runs the injector init container.
  - **postStart**: `exec` command symlinks the binary into PATH and optionally runs `setupCommand`.
- Two injection patterns: `init` (single binary copy) and `bundle` (full runtime directory with symlink).

Знімок екрана 2026-04-06 о 06 19 52
Знімок екрана 2026-04-06 о 06 21 04
Знімок екрана 2026-04-06 о 06 21 32
Знімок екрана 2026-04-06 о 06 22 08
Знімок екрана 2026-04-06 о 06 22 30
Знімок екрана 2026-04-06 о 06 22 59

fixes https://github.com/eclipse-che/che/issues/23796

#### Setting up the AI Tool Registry

The dashboard reads AI tool definitions from a Kubernetes ConfigMap at runtime. To enable the AI widgets you need to create this ConfigMap:

```bash
# Clone the AI tool images repository
git clone https://github.com/olexii4/che-ai-tool-images.git
cd che-ai-tool-images

# Set your Che namespace
NS="${CHE_NAMESPACE:-eclipse-che}"

# Create the AI tool registry ConfigMap
oc create configmap ai-tool-registry \
  --from-file=registry.json=registry.json \
  -n "$NS" \
  --dry-run=client -o yaml | \
  oc label --local -f - \
    app.kubernetes.io/component=ai-tool-registry \
    app.kubernetes.io/part-of=che.eclipse.org \
    -o yaml | \
  oc apply -f -
```

When no ConfigMap is found, the dashboard returns an empty registry and all AI-related UI elements are hidden automatically. To remove all AI tools:

```bash
oc delete configmap ai-tool-registry -n "${CHE_NAMESPACE:-eclipse-che}"
```

#### Manual testing steps

2. **Create Workspace page**: verify the AI Provider Selector accordion renders below the Editor Selector. Expand "Choose an AI Provider" — cards should show provider icons, names, and descriptions sorted alphabetically, navigable with arrow keys. Selecting a card and creating a workspace should inject the tool's init container, shared volume, and lifecycle commands into the DevWorkspace spec.
3. **Default provider message**: if the admin configures `defaultAiProvider` in the CheCluster CR, the "Use a Default AI Provider" panel should display the resolved provider name (e.g. "The default AI provider "Gemini" configured by your administrator will be used.").
4. **User Preferences → AI Providers Keys tab**: add an API key for a tool — verify the secret is created in the user namespace with correct labels (`che.eclipse.org/ai-provider-id`, `controller.devfile.io/mount-to-devworkspace`, `controller.devfile.io/watch-secret`) and annotation (`controller.devfile.io/mount-as: env`). Update and delete the key. Verify the "Key configured" badge updates on the Create Workspace gallery cards.
5. **Workspaces List**: verify the "AI Provider" column shows the injected tool icon and name for workspaces that have an AI tool, and "-" for those without.
6. **Workspace Details → Overview tab**: verify the AI Tool section shows the current tool or "None". Click the info icon — verify the modal shows tools with provider descriptions. Click edit, select a different tool, save — verify the DevWorkspace spec is updated. When no tools are configured, verify the empty-state message is shown.
7. **Workspace runtime**: start a workspace with an injected tool (e.g. Gemini CLI). Open a terminal in the editor container and verify the tool binary is available in PATH (e.g. `which gemini`). Verify the API key env var is set (e.g. `echo $GEMINI_API_KEY`).
8. **Empty registry**: delete the `ai-tool-registry` ConfigMap and verify that the AI Provider Selector on the Create Workspace page, the AI Provider(s) column in the Workspaces List, and the AI Providers Keys tab in User Preferences are all hidden.
9. **AI tool injection (Gemini CLI)**: on the Create Workspace page, select "Gemini CLI" as the AI Provider. Create a new workspace from any sample (e.g. Go). Wait for the workspace to start and the IDE to load. Open a terminal in the IDE and run `gemini` — verify the Gemini CLI starts (or prints its help/version output), confirming the AI tool binary was injected into PATH.
10. **AI tool injection (Claude Code)**: repeat step 9 but select "Claude Code" as the AI Provider. In the terminal, run `claude` — verify the Claude Code CLI is available.
11. **Multiple AI tools**: on the Create Workspace page, expand "Choose an AI Provider" and select multiple providers (e.g. both Gemini CLI and Claude Code). Create a workspace, open a terminal, and verify both `gemini` and `claude` commands are available. Check the Workspaces List — the "AI Provider(s)" column should show icons for both tools.

#### Docs PR

https://github.com/eclipse-che/che-docs/pull/3062


TheChosenMok (approved):
Looks good, only thing that comes to mind is adding on an additional verification method for credentials.json to support OAuth. Great Work!

packages/dashboard-frontend/src/components/AiSelector/index.tsx
```diff
@@ -0,0 +1,254 @@
...
+        ? prevState.selectedProviderIds.filter(id => id !== providerId)
+        : [...prevState.selectedProviderIds, providerId];
+
+      if (this.state.expandedId === 'selector') {
+        this.props.onSelect(selectedProviderIds);
+      }
+
+      return { selectedProviderIds };
+    });
+  }
```
akurinnoy: Two issues here:
1. side effect inside `setState()` updater
2. `this.state.expandedId` is read instead of using `prevState`

```suggestion
  private handleProviderToggle(providerId: string): void {
    this.setState((prevState) => {
      const isSelected = prevState.selectedProviderIds.includes(providerId);
      const selectedProviderIds = isSelected
        ? prevState.selectedProviderIds.filter(id => id !== providerId)
        : [...prevState.selectedProviderIds, providerId];

      return { selectedProviderIds };
    }, () => {
      if (this.state.expandedId === 'selector') {
        this.props.onSelect(this.state.selectedProviderIds);
    });
  }
```

packages/dashboard-frontend/src/services/backend-client/aiConfigApi.ts
```diff
@@ -0,0 +1,65 @@
...
+  toolId: string,
+  envVarName: string,
+  apiKey: string,
+): Promise<void> {
+  try {
+    const encodedKey = btoa(apiKey);
+    await AxiosWrapper.createToRetryMissedBearerTokenError().post(
+      `${dashboardBackendPrefix}/namespace/${namespace}/ai-provider-key`,
+      { toolId, envVarName, apiKey: encodedKey },
+    );
```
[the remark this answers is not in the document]
olexii4: I have fixed it

packages/dashboard-backend/src/routes/api/aiConfig.ts
```diff
@@ -0,0 +1,96 @@
...
+
+const tags = ['AI Config'];
+const rateLimitConfig = {
+  config: {
+    rateLimit: {
+      max: 100,
+      timeWindow: '1 minute',
+    },
+  },
+};
```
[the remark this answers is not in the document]
olexii4: The package @fastify/rate-limit is already a dependency on main — it was added for the air-gap sample routes (**airGapSample.ts**). This PR doesn't introduce it.

packages/dashboard-frontend/src/services/helpers/aiTools.ts
```diff
@@ -0,0 +1,726 @@
...
+      const pathEntry = editorComp.container.env.find(e => e.name === 'PATH');
+      if (pathEntry) {
+        if (!pathEntry.value.includes('/injected-tools/bin')) {
+          pathEntry.value = `/injected-tools/bin:${pathEntry.value}`;
+        }
+      } else {
+        editorComp.container.env.push({
+          name: 'PATH',
+          value: '/injected-tools/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
+        });
```
akurinnoy, line 322: Thanks for testing this! Makes sense — `$(PATH)` is resolved by the operator after container startup, so the entrypoint process sees the literal string `$(PATH)` as a path component and fails to find itself.

The current approach is acceptable given the constraint. Could you add a short comment near the fallback explaining why the hardcoded value is used and that it matches the che-code image's default PATH? Something like:

```ts
} else {
  // $(PATH) substitution cannot be used here — it is resolved by the operator
  // after container startup, causing entrypoint failures. The fallback value
  // matches the default PATH of the che-code editor image.
  editorComp.container.env.push({
    name: 'PATH',
    value: '/injected-tools/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
  });
}
```

packages/dashboard-frontend/src/services/helpers/aiTools.ts
```diff
@@ -0,0 +1,726 @@
...
+  // Collect slugs of all injector components that will remain after stale removal
+  const staleNames = new Set(staleComponents.map(c => c.name).filter(Boolean));
+  const remainingInjectorSlugs = new Set(
+    components
+      .filter(c => c.name?.endsWith('-injector') && !staleNames.has(c.name))
+      .map(c => c.name!.replace(/-injector$/, '')),
+  );
+
+  // Find orphaned tool commands whose injector component no longer exists.
+  // Cleanup commands are included — by the next workspace start they have already
```
akurinnoy: I looked at this more carefully and found a sequencing issue. When a tool is removed:

1. `removeAiToolFromWorkspace` adds a `cleanup-<slug>` postStart command and removes the `<slug>-injector` component from the spec.
2. On the very next `startWorkspace`, `sanitizeStaleAiTools` runs first. It sees `cleanup-<slug>` as an orphan (no matching `<slug>-injector` component) and **removes the cleanup command from the spec before the workspace boots**.
3. The workspace starts — `cleanup-<slug>` was never executed. The binary stays on the PVC.

The comment here ("by the next workspace start they have already executed") seems to assume the cleanup ran before `sanitizeStaleAiTools` removes it, but the order is the opposite on a cold start-after-removal. The cleanup only survives to a second start if the workspace was already running when the tool was removed (i.e. the cleanup ran in that session's background).

Not a blocker — the binary sitting on the PVC is harmless — but the stated design goal isn't achieved on the first cold start. Worth a follow-up issue if you want to track it.

packages/dashboard-backend/src/devworkspaceClient/services/aiRegistryApi.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    for (const cm of response.items) {
+      if (cm.data === undefined) {
+        continue;
+      }
+
+      for (const key in cm.data) {
+        try {
+          const parsed = JSON.parse(cm.data[key]) as Record<string, unknown>;
```
akurinnoy: Minor: the cast to `Record<string, unknown>` is unsafe if `JSON.parse` returns a non-object (e.g. a string or array — both valid JSON). I see that `.providers`, `.tools`, and `.defaultAiProviders` are already guarded with `Array.isArray` below, which handles `undefined` gracefully. The one gap is if `parsed` itself is not an object — accessing `.providers` on a primitive or array won't throw in JS but it will silently return `undefined`, which the fallbacks cover anyway.

So this is more of a type hygiene note than a real bug. A minimal guard:

```ts
const parsed = JSON.parse(cm.data[key]);
if (typeof parsed !== 'object' || parsed === null || Array.isArray(parsed)) {
  return acc;
}
```

Not a blocker given the existing fallbacks.

packages/dashboard-frontend/src/services/helpers/aiTools.ts
olexii4: I have added a three-line comment directly above the `env.push` call:

```typescript
// $(PATH) substitution cannot be used here -- it is resolved by the operator
// after container startup, causing entrypoint failures. The fallback value
// matches the default PATH of the che-code editor image.
```

packages/dashboard-backend/src/devworkspaceClient/services/aiRegistryApi.ts
olexii4: Thanks. I have added a runtime type guard before casting:

```typescript
const parsed: unknown = JSON.parse(cm.data[key]);
if (typeof parsed !== 'object' || parsed === null || Array.isArray(parsed)) {
  continue;
}
const registry = parsed as Record<string, unknown>;
```

This skips non-object values (primitives, null, arrays) and only proceeds when the parsed value is a plain object.

packages/dashboard-frontend/src/services/helpers/aiTools.ts
olexii4: Good catch, fixed. The cleanup command was indeed removed before it could execute on a cold start.

The fix uses annotation-based state tracking via `che.eclipse.org/pending-ai-cleanup`:

1. `removeAiToolFromWorkspace` adds the cleanup command **and** sets the annotation with the tool slug.
2. On the first cold start, `sanitizeStaleAiTools` checks the annotation:
   - Slug is in the annotation -> cleanup has **not** executed yet -> **keep** the command, **clear** the slug from the annotation.
3. On the second cold start, the annotation is gone:
   - No pending slug -> cleanup has already executed -> **remove** the command.

**Why not the simpler regex-only fix?**

Excluding `cleanup` from the orphan detection regex (`/^(install|symlink|run)-(.+)$/`) would let cleanup commands survive one start, but they would **never** be removed afterward. A cleanup command without an injector is always an orphan by the regex definition, so it would persist indefinitely. The annotation approach gives exactly one-start-cycle survival.

akurinnoy (approved):
Great job!

ibuziuk (commented):
Please, do not merge for 7.117 / 3.28 
We need more follow up on that flow till it hit upstream release

ibuziuk:
@olexii4 please, update the PR to resolve the conflicts - the plan is to have the PR merged in 7.118

ibuziuk (commented):
@olexii4 let's add the Tech preview label to the widget, similar to what we have on devfile samples:
Image

dmytro-ndp:
I have a few questions:
1) there are also codex CLI and Copilot CLI available in the Market. Should we support them as well?
2) do you need AI CLI owner agreements to place their tile in the User Dashboard?
3) will there be team wide API keys supported which covers several accounts?

olexii4:
> I have a few questions:
>
> 1. There are also Codex CLI and Copilot CLI available in the market. Should we support them as well?
> 2. Do you need AI CLI owner agreements to place their tile in the User Dashboard?
> 3. Will there be team-wide API keys supported which cover several accounts?

@dmytro-ndp, here is my take on this:

**Q1 — Should we support Codex CLI, Copilot CLI, or other tools?**

This PR introduces only the **injection mechanism** — it does not ship, bundle, or endorse any specific AI CLI. The responsibility for which tools are available rests entirely with the **cluster administrator** who deploys the `ai-tool-registry` ConfigMap. The Dashboard simply reads whatever providers and tools the admin has registered and renders them.

Any CLI can be supported — Codex CLI, Copilot CLI, or anything else — as long as an admin publishes a ConfigMap with the corresponding provider/tool entry and a container image that performs the injection. The current reference registry at [github.com/olexii4/che-ai-tool-images](https://github.com/olexii4/che-ai-tool-images) (Gemini, Claude Code, OpenCode) exists solely to validate the mechanism and is not meant to be exhaustive.

---

**Q2 — Do we need agreements with AI CLI owners to show their tile in the Dashboard?**

No formal agreement is required on Che's side, for two reasons:

1. **Che does not distribute the AI software.** The CLI binary lives in a container image the admin supplies — Che only reads metadata (name, icon URL, description) from a ConfigMap. There is no bundling, redistribution, or sublicensing of any vendor's software by Eclipse Che itself.

2. **Icons are MIT-licensed.** The current registry uses the [lobehub open-source icon pack](https://github.com/lobehub/lobe-icons), which covers Gemini, Claude, and similar marks for product-identification purposes in developer tools — standard nominative fair use.

That said, **administrators** adding a specific tool to their registry take on the ToS responsibility for that tool. A couple of things worth flagging:

- **Anthropic** — their 2025 commercial terms restrict third-party tools from spoofing the official Claude Code client to obtain more favourable API limits. Admins deploying a custom Claude image should ensure it complies. ([Anthropic Commercial Terms](https://www.anthropic.com/legal/commercial-terms))
- **GitHub Copilot CLI** — governed by a non-transferable, limited licence. An org wishing to inject it would need a valid Copilot Business/Enterprise subscription and should confirm that injecting it into shared containers falls within their licence. ([GitHub Copilot Policy](https://docs.github.com/en/site-policy/github-terms/github-copilot-extension-developer-policy))

---

**Q3 — Will team-wide API keys covering several accounts be supported?**

Not in this implementation — and doing it properly is non-trivial.

The current design injects the API key as a **per-workspace environment variable** sourced from a per-user Kubernetes Secret. Each user manages their own key. A shared team key would require a cluster-level Secret readable across user namespaces, which creates real problems:

- Any user could trivially extract it via `echo $ANTHROPIC_API_KEY`.
- Cost and quota attribution across users becomes opaque.
- One compromised workspace exposes the key for everyone.

Beyond that, AI providers issue keys at the individual or organisation level and offer no built-in per-seat partitioning, so tracking who consumed what is impossible without an intermediary.

The right path for team-wide keys would be a **proxy gateway** — something like [LiteLLM](https://www.litellm.ai/) — that holds the master key server-side, enforces per-user quotas, and hands short-lived scoped tokens to individual workspaces. That is a separate, substantial feature and out of scope for this PR.

---

@ibuziuk WDYT?

olexii4:
> @olexii4 let's add the Tech preview label to the widget, similar to what we have on devfile samples: Image

@ibuziuk  I have fixed it:

Знімок екрана 2026-05-09 о 23 49 47

Знімок екрана 2026-05-09 о 23 49 24

svor (approved):
lgtm
Screenshot 2026-06-01 at 15 34 16
Screenshot 2026-06-01 at 15 33 58

PR #1530, opened 2026-04-20
fix: correct editor binary detection and always create new Empty Workspace

### LINKED ISSUE #23798: 'Downloading IDE binaries' is displayed for Che-Code editor
### Describe the bug

`Downloading IDE binaries` is displayed for Che-Code editor 

### Che version

7.116@latest

### Steps to reproduce

Start a workspace using this link: https://che-dogfooding.apps.che-dev.x6e0.p1.openshiftapps.com/#git@github.com:RomanNikitenko/che-code.git

### Expected behavior

`Downloading IDE binaries` should not be displayed for Che-Code editor 

### Screenshots

Image

### Environment

Dogfooding instance

olexii4:
### What does this PR do?
This PR contains two fixes:

1. **Use allow-list for editors with binary downloads** — The "Downloading IDE binaries" step was incorrectly shown for Che-Code and other editors that do not download binaries. The previous implementation used a deny-list (`EDITORS_WITHOUT_BINARIES`) that only excluded `che-code`, causing false positives for any new editor not explicitly listed. This is replaced with an allow-list (`EDITORS_WITH_BINARIES`) that explicitly enumerates the JetBrains-based editors that actually download binaries: `che-idea-server`, `che-clion-server`, `che-phpstorm-server`, `che-pycharm-server`, `che-rider-server`, `che-rubymine-server`, `che-webstorm-server`, `che-goland-server`.

2. **Always create new workspace for Empty Workspace sample** — When clicking "Create Empty Workspace", the dashboard would reuse an existing workspace if one had been created from the same empty devfile (due to the default `policies.create=peruser` deduplication). This fix forces `policies.create=perclick` for any sample tagged `"Empty"`, ensuring a new workspace is always created.

fixes https://github.com/eclipse-che/che/issues/23798
fixes https://redhat.atlassian.net/browse/CRW-10546

### Is it tested? How?

**Editor binary detection:**
1. Deploy Eclipse Che with the image from this PR.
2. Start a workspace using **Che-Code** editor (e.g., `che-incubator/che-code/latest`).
3. Verify that the "Downloading IDE binaries" step is **not** shown during workspace startup.
4. Start a workspace using a **JetBrains** editor (e.g., `che-incubator/che-idea-server/latest`).
5. Verify that the "**Downloading IDE binaries**" step **is** shown during workspace startup.

**Empty Workspace creation:**
1. Deploy Eclipse Che with the image from this PR.
2. Unselect "Create New" button (default is "on").
3. Click "**Create Empty Workspace**" from the Get Started page.
4. Wait for the workspace to start.
5. Click "**Create Empty Workspace**" again.
6. Verify that a **new** workspace is created (not redirected to the existing one).

### DIFF
--- a/packages/dashboard-frontend/src/components/WorkspaceProgress/StartingSteps/WorkspaceConditions/helpers.ts
+++ b/packages/dashboard-frontend/src/components/WorkspaceProgress/StartingSteps/WorkspaceConditions/helpers.ts
@@ -10,20 +10,25 @@
  *   Red Hat, Inc. - initial API and implementation
  */
 
+import { load } from 'js-yaml';
+
 import { WorkspaceRouteParams } from '@/Routes';
 import { DEVWORKSPACE_CHE_EDITOR } from '@/services/devfileApi/devWorkspace/metadata';
 import { Workspace } from '@/services/workspace-adapter';
 
-export const EDITORS_WITH_BINARIES = [
-  'che-idea-server',
-  'che-clion-server',
-  'che-phpstorm-server',
-  'che-pycharm-server',
-  'che-rider-server',
-  'che-rubymine-server',
-  'che-webstorm-server',
-  'che-goland-server',
-];
+export const EDITORS_WITHOUT_BINARIES = ['che-code'];
+
+function getEditorName(cheEditor: string): string | undefined {
+  if (cheEditor.includes('\n')) {
+    try {
+      const parsed = load(cheEditor) as { metadata?: { name?: string } };
+      return parsed?.metadata?.name;
+    } catch {
+      return undefined;
+    }
+  }
+  return cheEditor.split('/')[1];
+}
 
 export function hasDownloadBinaries(
   allWorkspaces: Workspace[],
@@ -40,9 +45,12 @@ export function hasDownloadBinaries(
     return false;
   }
   const cheEditor = targetWorkspace.ref.metadata.annotations?.[DEVWORKSPACE_CHE_EDITOR];
-  if (!cheEditor || cheEditor.startsWith('apiVersion') || cheEditor.startsWith('schemaVersion')) {
+  if (!cheEditor) {
     return false;
   }
-  const name = cheEditor.split('/')[1];
-  return EDITORS_WITH_BINARIES.includes(name);
+  const name = getEditorName(cheEditor);
+  if (!name) {
+    return false;
+  }
+  return !EDITORS_WITHOUT_BINARIES.includes(name);
 }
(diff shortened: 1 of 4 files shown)


packages/dashboard-frontend/src/components/WorkspaceProgress/StartingSteps/WorkspaceConditions/helpers.ts
```diff
@@ -14,32 +14,35 @@ import { WorkspaceRouteParams } from '@/Routes';
 import { DEVWORKSPACE_CHE_EDITOR } from '@/services/devfileApi/devWorkspace/metadata';
 import { Workspace } from '@/services/workspace-adapter';
 
-export const EDITORS_WITHOUT_BINARIES = ['che-code'];
```
svor, line 17: @olexii4 could you explain why this approach doesn't work?

packages/dashboard-frontend/src/components/WorkspaceProgress/StartingSteps/WorkspaceConditions/helpers.ts
```diff
@@ -14,32 +14,35 @@ import { WorkspaceRouteParams } from '@/Routes';
...
+export const EDITORS_WITH_BINARIES = [
+  'che-idea-server',
```
svor: as I commended here: https://github.com/eclipse-che/che-dashboard/pull/1405#pullrequestreview-3428918507, I don't like this big hardcoded list, but if we don't have another solution, should we add `che-code-server` https://github.com/eclipse-che/che-operator/blob/main/editors-definitions/che-code-server-latest.yaml into the list?

packages/dashboard-frontend/src/components/WorkspaceProgress/StartingSteps/WorkspaceConditions/helpers.ts
olexii4: The `che.eclipse.org/che-editor` annotation on a DevWorkspace can contain **two different formats**:

#### Format 1 — Short Editor ID

When the dashboard resolves the editor from its registry:

```
che-incubator/che-code/insiders
```

In this case, `cheEditor.split('/')[1]` correctly extracts `"che-code"`, and the check works.

#### Format 2 — Full Devfile YAML Content

When a repository provides a custom editor via `.che/che-editor.yaml` (e.g.,
[RomanNikitenko/che-code/.che/che-editor.yaml](https://github.com/RomanNikitenko/che-code/blob/main/.che/che-editor.yaml)
which contains `id: che-incubator/che-code/insiders`), the dashboard resolves the editor ID to its
full devfile content and stores the **entire devfile YAML** in the annotation:

```yaml
schemaVersion: 2.3.0
metadata:
  name: che-code
components:
  - container:
      command:
        - /entrypoint-init-container.sh
      image: quay.io/che-incubator/che-code:insiders
    name: che-code-injector
  ...
```

#### Why the Guard Fails

Implementation:

```typescript
if (!cheEditor || cheEditor.startsWith('apiVersion') || cheEditor.startsWith('schemaVersion')) {
  return false;
}
```

packages/dashboard-frontend/src/components/WorkspaceProgress/StartingSteps/WorkspaceConditions/helpers.ts
olexii4: I have fixed it

svor (approved):
@olexii4 thanks, works as expected!

Image

Image

PR #1538, opened 2026-04-21
fix: add arrow key navigation to User Preferences tabs

olexii4:
### What does this PR do?
The tab list on the User Preferences page (Container Registries, Git Services, Personal Access Tokens, Gitconfig, SSH Keys) did not follow WAI-ARIA keyboard navigation patterns. Arrow keys did not switch between tabs, and pressing Tab moved focus to the next tab header instead of into the active tab's content panel.

PatternFly 5's `Tabs` component does not implement built-in arrow key navigation. This PR adds an `onKeyDown` handler to the `Tabs` component that implements the WAI-ARIA tabs keyboard pattern:
- **Arrow Left / Arrow Right** -- switches focus and activates the previous/next tab (wraps around)
- **Arrow Up / Arrow Down** -- same behavior for vertical navigation
- **Home / End** -- jumps to the first/last tab

fixes https://redhat.atlassian.net/browse/CRW-10288

### Is it tested? How?

1. Deploy Eclipse Che with the image from this PR.
2. Navigate to the **User Preferences** page
3. Use **Tab** to move focus to the first tab header (Container Registries)
4. Press **Arrow Right** -- focus should move to "Git Services" and the tab panel should switch
5. Press **Arrow Right** repeatedly -- focus should cycle through all tabs and wrap around
6. Press **Arrow Left** -- focus should move to the previous tab
7. Press **Home** -- focus should jump to the first tab (Container Registries)
8. Press **End** -- focus should jump to the last tab (SSH Keys)

### DIFF
--- a/packages/dashboard-frontend/src/pages/UserPreferences/index.tsx
+++ b/packages/dashboard-frontend/src/pages/UserPreferences/index.tsx
@@ -37,6 +37,14 @@ export type State = {
 };
 
 class UserPreferences extends React.PureComponent<Props, State> {
+  private readonly tabOrder: UserPreferencesTab[] = [
+    UserPreferencesTab.CONTAINER_REGISTRIES,
+    UserPreferencesTab.GIT_SERVICES,
+    UserPreferencesTab.PERSONAL_ACCESS_TOKENS,
+    UserPreferencesTab.GITCONFIG,
+    UserPreferencesTab.SSH_KEYS,
+  ];
+
   constructor(props: Props) {
     super(props);
 
@@ -68,17 +76,49 @@ class UserPreferences extends React.PureComponent<Props, State> {
     return UserPreferencesTab.CONTAINER_REGISTRIES;
   }
 
-  private handleTabClick(
+  private handleTabClick = (
     event: React.MouseEvent<HTMLElement, MouseEvent>,
     activeTabKey: string | number,
-  ): void {
+  ): void => {
     event.stopPropagation();
     this.props.navigate(`${ROUTE.USER_PREFERENCES}?tab=${activeTabKey}`);
 
     this.setState({
       activeTabKey: activeTabKey as UserPreferencesTab,
     });
-  }
+  };
+
+  private handleTabKeyDown = (event: React.KeyboardEvent): void => {
+    const target = event.target as HTMLElement;
+    if (target.getAttribute('role') !== 'tab') {
+      return;
+    }
+
+    const { activeTabKey } = this.state;
+    const currentIndex = this.tabOrder.indexOf(activeTabKey);
+    let nextIndex = -1;
+
+    if (event.key === 'ArrowRight' || event.key === 'ArrowDown') {
+      nextIndex = (currentIndex + 1) % this.tabOrder.length;
+    } else if (event.key === 'ArrowLeft' || event.key === 'ArrowUp') {
+      nextIndex = (currentIndex - 1 + this.tabOrder.length) % this.tabOrder.length;
+    } else if (event.key === 'Home') {
+      nextIndex = 0;
+    } else if (event.key === 'End') {
+      nextIndex = this.tabOrder.length - 1;
+    } else {
+      return;
+    }
+
+    event.preventDefault();
+    const nextTabKey = this.tabOrder[nextIndex];
+    this.props.navigate(`${ROUTE.USER_PREFERENCES}?tab=${nextTabKey}`);
+    this.setState({ activeTabKey: nextTabKey }, () => {
+      const tabsContainer = document.getElementById('user-preferences-tabs');
+      const buttons = tabsContainer?.querySelectorAll<HTMLButtonElement>('[role="tab"]');
+      buttons?.[nextIndex]?.focus();
+    });
+  };
 
   render(): React.ReactNode {
     const { activeTabKey } = this.state;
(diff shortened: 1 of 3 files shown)


olexii4:
@sskoryk Could you review this PR?

svor:
@SkorikSergey we need your review

SkorikSergey (approved):
Looks good to merge

packages/dashboard-frontend/src/pages/UserPreferences/index.tsx
```diff
@@ -47,6 +55,27 @@ class UserPreferences extends React.PureComponent<Props, State> {
...
+  private updateTabIndexes(): void {
+    const { activeTabKey } = this.state;
+    for (const tabKey of this.tabOrder) {
+      const tabButtonId = `pf-tab-${tabKey}-user-preferences-tabs`;
+      const button = document.getElementById(tabButtonId);
+      if (button) {
+        button.tabIndex = tabKey === activeTabKey ? 0 : -1;
+      }
+    }
+  }
```
akurinnoy: The `updateTabIndexes()` method uses `document.getElementById()` to imperatively set `tabIndex` on PatternFly tab buttons. This approach bypasses React's virtual DOM and has a few risks:

- If PatternFly re-renders the tab buttons for any reason, the tabIndex values will be silently overwritten
- Creates timing dependencies with `componentDidMount` and `mountOnEnter={true}`
- Relies on PatternFly's internal DOM ID pattern (see [related comment](https://github.com/eclipse-che/che-dashboard/pull/1538/changes#r3137683522)), which is an implementation detail that can change

PatternFly 6's `Tab` component extends `React.HTMLProps<HTMLButtonElement>`, so `tabIndex` can be passed declaratively as a prop on each `<Tab>`:

```tsx
<Tab
  eventKey={UserPreferencesTab.CONTAINER_REGISTRIES}
  title="Container Registries"
  tabIndex={activeTabKey === UserPreferencesTab.CONTAINER_REGISTRIES ? 0 : -1}
>
```

This would eliminate the lifecycle methods for tabIndex management and the dependency on internal PF6 DOM IDs.

packages/dashboard-frontend/src/pages/UserPreferences/index.tsx
```diff
@@ -80,6 +109,37 @@ class UserPreferences extends React.PureComponent<Props, State> {
...
+    } else {
+      return;
+    }
+
+    event.preventDefault();
+    const nextTabKey = this.tabOrder[nextIndex];
+    this.props.navigate(`${ROUTE.USER_PREFERENCES}?tab=${nextTabKey}`);
+    this.setState({ activeTabKey: nextTabKey }, () => {
+      const tabButtonId = `pf-tab-${nextTabKey}-user-preferences-tabs`;
+      document.getElementById(tabButtonId)?.focus();
```
akurinnoy: The `setState` callback calls `focus()` immediately after the state update. Since `componentDidUpdate` (which calls `updateTabIndexes()`) fires asynchronously after render, the focus call may run before the tabIndex values are updated on the new tab button.

If we switch to declarative `tabIndex` props as suggested in the comment above, this timing issue would resolve naturally.

packages/dashboard-frontend/src/pages/UserPreferences/index.tsx
```diff
@@ -47,6 +55,27 @@ class UserPreferences extends React.PureComponent<Props, State> {
...
+    if (prevState.activeTabKey !== this.state.activeTabKey) {
+      this.updateTabIndexes();
+    }
+  }
+
+  private updateTabIndexes(): void {
+    const { activeTabKey } = this.state;
+    for (const tabKey of this.tabOrder) {
+      const tabButtonId = `pf-tab-${tabKey}-user-preferences-tabs`;
+      const button = document.getElementById(tabButtonId);
```
akurinnoy: The tab button ID format `pf-tab-${tabKey}-user-preferences-tabs` mirrors PatternFly 6's internal ID generation pattern (`pf-tab-${eventKey}-${tabsId}`). This is an undocumented implementation detail - it could silently break on PF6 upgrades that change the ID format.

Consider adding a comment to make this dependency explicit:

```suggestion
      // Note: tab button IDs follow PF6's internal pattern: pf-tab-${eventKey}-${tabsId}
      // This format is not part of PF6's public API and may change across versions
      const tabButtonId = `pf-tab-${tabKey}-user-preferences-tabs`;
      const button = document.getElementById(tabButtonId);
```

packages/dashboard-frontend/src/pages/UserPreferences/index.tsx
```diff
@@ -80,6 +109,37 @@ class UserPreferences extends React.PureComponent<Props, State> {
     });
   }
 
+  private handleTabKeyDown = (event: React.KeyboardEvent): void => {
```
akurinnoy: I'm curious why do you mix class methods and fields. Was it done deliberately, for a purpose?
Maybe, update other event handlers this way, for consistency?

olexii4: I agree it's inconsistent. I have fixed it.

packages/dashboard-frontend/src/pages/UserPreferences/index.tsx
olexii4: I have fixed it

packages/dashboard-frontend/src/pages/UserPreferences/index.tsx
olexii4: I have fixed it

akurinnoy (approved):

packages/dashboard-frontend/src/pages/UserPreferences/index.tsx
olexii4: I have fixed it

PR #1548, opened 2026-04-30
fix(ui): disable Save button when Gitconfig name or email fields are empty

vinokurig:
### What does this PR do?
Initialize validation state to false instead of true in `GitConfigSectionUser` to ensure the Save button remains disabled when one of the field is empty, preventing users from saving invalid gitconfig data.

https://redhat.atlassian.net/browse/CRW-10719

### Is it tested? How?

1. Go to `User Preferences` GitConfig tab.
2. Enter a valid username to the `Name` field.
3. See: The `Save` button is disabled.


olexii4 (approved):
LGTM

olkornii (approved):
Works for me with 3.27.1 and quay.io/eclipse/che-dashboard:pr-1548.

PR #1554, opened 2026-05-05
fix(a11y): replace nested anchor in button with onClick handler

vinokurig:
### What does this PR do?
Add the `onClick` handler to the `open registry` button instead of the anchor element in order to get rid of the nested `<a>` element.

https://redhat.atlassian.net/browse/CRW-10317

### Is it tested? How?

1. Navigate the browser's element selector to the `Open Registry` icon.
2. See: no nested `<a>` element is present in the button structure:
image


olexii4 (approved):
LGTM

dmytro-ndp (approved):
@vinokurig : I checked the PR using Eclipse Che Next with the User Dashboard from quay.io/eclipse/che-dashboard:pr-1554

There was no nested anchor in the button near the "Registry" field of the "Add Container Registry" HTML form, which means the accessibility issue is fixed properly:
Image

At the same time, the button worked as expected—it redirected to the registry via the URL from the field.

Well done!

PR #1563, opened 2026-05-12
fix(a11y): add role="img" to workspace status indicator span (WCAG 4.1.2)

olexii4:
### What does this PR do?
Fixes [WCAG 2.2 criterion 4.1.2 Name, Role, Value (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html) for the workspace status indicator icon.

The `WorkspaceStatusIndicator` component renders a `<span>` with an `aria-label` (e.g. `"Workspace status is Running"`) but no semantic role. Per the ARIA specification, naming attributes such as `aria-label` are only honoured by assistive technologies when the element has a valid role. On a generic `<span>`, screen readers silently ignore the label — users who rely on a screen reader receive no information about the workspace status.

This PR adds `role="img"` to the span, declaring it as a named graphic element. The `aria-label` is now correctly exposed to the accessibility API and announced by screen readers when the element receives focus.

**Affected surfaces:** Workspaces list table (status column) and Workspace Details Overview tab.

```diff
 <span
+  role="img"
   className={styles.statusIndicator}
   data-testid="workspace-status-indicator"
   aria-label={`Workspace status is ${status}`}
 >
   {icon}
 </span>
```

The `role="img"` value is the correct choice here: the element contains a purely visual icon that conveys information, and the `aria-label` supplies its accessible text alternative — exactly the pattern described in [ARIA authoring practices for images](https://www.w3.org/WAI/ARIA/apg/patterns/img/).

fixes https://redhat.atlassian.net/browse/CRW-10233

### Is it tested? How?

- 16 snapshots updated in `Indicator.spec.tsx.snap` to include `role="img"`.
- Manually verified with macOS VoiceOver on Chrome: the status label is now announced on focus for both the Workspaces list and the Workspace Details Overview.

### DIFF
--- a/packages/dashboard-frontend/src/components/Workspace/Status/Indicator/__tests__/__snapshots__/Indicator.spec.tsx.snap
+++ b/packages/dashboard-frontend/src/components/Workspace/Status/Indicator/__tests__/__snapshots__/Indicator.spec.tsx.snap
@@ -21,6 +21,7 @@ exports[`Workspace indicator component Che Workspaces should render ERROR status
       aria-label="Workspace status is ERROR"
       className="statusIndicator"
       data-testid="workspace-status-indicator"
+      role="img"
     >
       <span
         className="pf-v6-c-icon pf-m-inline"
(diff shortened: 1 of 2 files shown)


svor (approved):
lgtm

dmytro-ndp (approved):
@olexii4 : I tested Eclipse Che Next with the User Dashboard image quay.io/eclipse/che-dashboard:pr-1563, and the correct HTML attribute `role="img"` was added as expected:

Image

akurinnoy (approved):

PR #1567, opened 2026-05-14
fix: minimal fix for airgap sample trust and factory URL source comparison

olexii4:
### What does this PR do?

Fixes two related issues that caused the "Create New" workspace policy to be ignored
when creating workspaces from devfile registry samples or airgap samples:

1. **Airgap sample trust gate** — airgap sample URLs were not automatically trusted,
   causing the untrusted-source modal to block the factory flow.

2. **Source URL comparison mismatch** — `CheckExistingWorkspaces` compares
   `workspace.source` (read from the DevWorkspace annotation) against
   `factoryParams.sourceUrl` (derived from the current factory URL). Two bugs caused
   these values to differ, so the step always created a new workspace regardless of
   the "Create New" setting:

   - **Double-encoding in airgap URLs.** The factory hash encodes the inner devfile
     download URL twice (`%253F` → `%3F` after the first `URLSearchParams` decode).
     `workspace.source` applied a second decode via `URLSearchParams.get()` (`%3F` → `?`),
     while `factoryParams.sourceUrl` kept the single-encoded form. Result:
     ```
     factoryParams.sourceUrl → "…download%3Fid%3Dnodejs-express"
     workspace.source        → "…download?id=nodejs-express"   ← mismatch
     ```

   - **Value truncation on `=` in URL parameters.** The old annotation parser used
     `split('&')` + `split('=')[1]` to extract the `url` value. For any URL whose
     query string contains an `=` sign (e.g. `?id=nodejs-express`), `split('=')[1]`
     returned only the fragment up to the second `=`, silently truncating the value.

fixes https://redhat.atlassian.net/browse/CRW-9659

### Is it tested? How?

  1. Deploy OpenShift Dev Spaces with changes from this PR.
  2. Create a workspace from a sample with "Create New" OFF.
  3. Navigate to the same factory URL — existing workspace is opened, no duplicate.
  4. Repeat with an **airgap** sample — same behaviour.

### DIFF
--- a/.deps/EXCLUDED/dev.md
+++ b/.deps/EXCLUDED/dev.md
@@ -5,6 +5,7 @@ This file contains a manual contribution to .deps/dev.md and it's needed because
 | `@eclipse-che/api@7.86.0` | ecd.che |
 | `@eclipse-che/license-tool@2.0.0` | ecd.che |
 | `axios@1.15.2` | [clearlydefined](https://clearlydefined.io/definitions/npm/npmjs/-/axios/1.15.2) |
+| `jsbn@0.1.1` | [clearlydefined](https://clearlydefined.io/definitions/npm/npmjs/-/jsbn/0.1.1) |
 | `postcss@8.5.13` | [clearlydefined](https://clearlydefined.io/definitions/npm/npmjs/-/postcss/8.5.13) |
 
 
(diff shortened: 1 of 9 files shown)


svor:
@SkorikSergey could you please check the fix with DS image **quay.io/redhat-user-workloads/devspaces-tenant/devspaces/dashboard-rhel9:on-pr-6cf673cd9af66b5d337ea87b32d61759a8e3ee28**

SkorikSergey:
> @SkorikSergey could you please check the fix with DS image **quay.io/redhat-user-workloads/devspaces-tenant/devspaces/dashboard-rhel9:on-pr-6cf673cd9af66b5d337ea87b32d61759a8e3ee28**

Works as expected on "online" DS 3.28.0 with this PR dashboard image. With factory URLs and samples.

svor (approved):

SkorikSergey (approved):

PR #1571, opened 2026-05-18
fix(a11y): add inert to collapsed sidebar to remove focusable elements

olexii4:
### What does this PR do?
Fixes a WCAG 4.1.2 (Name, Role, Value — Level A) violation where the collapsed
left sidebar (`#page-sidebar`) was marked `aria-hidden="true"` by PatternFly but
still contained fully focusable `<a>` elements (e.g. Recent Workspaces links).

Screen reader users could silently tab into invisible interactive elements,
receiving no feedback from their assistive technology — a broken and confusing
navigation experience.

### Root Cause

PatternFly v6's `PageSidebar` sets `aria-hidden={!isSidebarOpen}` on its root
`<div>` when collapsed, but does not prevent child elements from remaining in the
keyboard tab order.  The ARIA spec requires that elements with `aria-hidden="true"`
are not focusable or reachable via keyboard; PatternFly delegates this
responsibility to the application.

### Fix

The `Sidebar` component now reads `isSidebarOpen` from PatternFly's `PageContext`
(which is provided by the parent `Page` component) and conditionally spreads the
HTML `inert` attribute onto `PageSidebar` when the sidebar is collapsed:

```tsx
const { isSidebarOpen } = React.useContext(PageContext);
const inertProp = isSidebarOpen ? {} : ({ inert: '' } as React.HTMLAttributes<HTMLDivElement>);

return (
  <PageSidebar {...inertProp}>
    <PageSidebarBody>
      <Navigation history={history} />
    </PageSidebarBody>
  </PageSidebar>
);
```

The HTML `inert` attribute removes the **entire subtree** from:
- Keyboard navigation (tab order)
- Pointer events
- The accessibility tree

This satisfies both remediation options from the bug report:
- **Option 1** — interactive children are removed from the tab order when collapsed
- **Option 2** — effectively equivalent to a proper CSS hide for AT purposes,
  without unmounting the component

fixes https://redhat.atlassian.net/browse/CRW-10739

### Is it tested? How?

2. Open the dashboard.
3. Click the hamburger button (top-left) to collapse the left sidebar.
4. Press `Tab` repeatedly — verify that focus never enters the collapsed sidebar
   (no Recent Workspaces links receive focus while the sidebar is hidden).
5. Expand the sidebar again — verify all navigation links are reachable with `Tab`
   as expected.

### DIFF
--- a/packages/dashboard-frontend/src/Layout/Sidebar/__tests__/__snapshots__/index.spec.tsx.snap
+++ b/packages/dashboard-frontend/src/Layout/Sidebar/__tests__/__snapshots__/index.spec.tsx.snap
@@ -1,6 +1,23 @@
 // Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
 
-exports[`Sidebar snapshot 1`] = `
+exports[`Sidebar snapshot - closed 1`] = `
+<div
+  aria-hidden={false}
+  className="pf-v6-c-page__sidebar pf-m-expanded"
+  id="page-sidebar"
+  inert=""
+>
+  <div
+    className="pf-v6-c-page__sidebar-body"
+  >
+    <div>
+      Mock Navigation component
+    </div>
+  </div>
+</div>
+`;
+
+exports[`Sidebar snapshot - open 1`] = `
 <div
   aria-hidden={false}
   className="pf-v6-c-page__sidebar pf-m-expanded"
--- a/packages/dashboard-frontend/src/Layout/Sidebar/__tests__/index.spec.tsx
+++ b/packages/dashboard-frontend/src/Layout/Sidebar/__tests__/index.spec.tsx
@@ -10,6 +10,8 @@
  *   Red Hat, Inc. - initial API and implementation
  */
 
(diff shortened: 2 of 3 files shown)


olexii4:
The **inert** global attribute is a Boolean attribute indicating that the element and all of its flat tree descendants become inert. The inert attribute can be added to sections of content that should not be interactive. When an element is inert, it along with all of the element's descendants, including normally interactive elements such as links, buttons, and form controls are disabled because they cannot receive focus or be clicked. The inert attribute can also be added to elements that should be offscreen or hidden. An **inert** element, along with its descendants, gets removed from the tab order and accessibility tree.

dmytro-ndp (approved):
@olexii4 : I checked Eclipse Che Next with User Dashboard quay.io/eclipse/che-dashboard:pr-1571 and can confirm that focus never enters the collapsed left sidebar when use Tab. 
At the same time expanded sidebar was reachable with Tab as expected:

https://github.com/user-attachments/assets/47ab4a71-a323-49db-9c29-34ce21e3fbdf

Good job!

svor (approved):

PR #1572, opened 2026-05-19
fix(a11y): prevent empty alert-group list from appearing in DOM

olexii4:
### What does this PR do?

Fixes a WCAG 4.1.2 (Name, Role, Value — Level A) violation where an empty
`<ul role="list" class="pf-v6-c-alert-group pf-m-toast">` element was briefly
present in the DOM without any `<li>` children, triggering the axe
`aria-required-children` rule.

### Root Cause

PatternFly v6's `AlertGroup` with `isToast={true}` manages its portal container
(a `<div>` appended to `document.body`) via a `useEffect` hook. In React's
execution order:

1. **Commit phase (synchronous, before paint):** React reconciles the component
   tree. When `AlertGroup` is unmounted (e.g. when all alerts are cleared), React
   synchronously removes the React-rendered content from the portal container —
   the `<ul>` is emptied.

2. **Browser paint:** The portal container `<div>` is still in `document.body`
   but its `<ul>` child is now empty (React has cleared it).

3. **Passive effects (after paint):** The `useEffect` cleanup runs and removes
   the portal container from `document.body`.

During the window between steps 1 and 3, an empty
`<ul role="list" class="pf-v6-c-alert-group pf-m-toast">` is in the DOM.
Automated accessibility tools (axe) can catch this window and report
`aria-required-children: list must contain listitem`.

This affected two components:

- **`AppAlertGroup`** — always-present global toast alert container
- **`ProgressAlert`** (WorkspaceProgress/Alert) — per-step toast alerts shown
  when the user is on the Logs or Events tab of the Workspace Creation page

### Fix

Replace `AlertGroup isToast` with `ReactDOM.createPortal` called directly in
`render()`. This change:

- **Eliminates the async gap**: `ReactDOM.createPortal` content is added and
  removed synchronously during React's commit phase (before paint), not via
  `useEffect`. The empty `<ul>` window no longer exists.
- **Preserves visual behaviour**: The `pf-m-toast` class is passed explicitly
  as `className` to `AlertGroup`, giving the `<ul>` the same CSS as before
  (`pf-v6-c-alert-group pf-m-toast`). Toast positioning is unchanged.
- **Preserves the existing guard**: Both components already return `<></>` when
  the alert list is empty, so `AlertGroup` (and its `<ul>`) is only rendered
  when there is at least one alert. The fix makes the guard effective even
  during teardown.

```tsx
// Before — async portal cleanup leaves empty <ul> briefly
return <AlertGroup isToast>{alerts}</AlertGroup>;

// After — synchronous portal management, <ul> never empty
return ReactDOM.createPortal(
  <AlertGroup className="pf-m-toast">{alerts}</AlertGroup>,
  document.body,
);
```

Fixes https://redhat.atlassian.net/browse/CRW-10740
### Is it tested? How?

2. Start a workspace to open the Workspace Creation page.
3. Navigate to the **Logs** tab.
4. Verify that no `aria-required-children` violation is reported for
   `.pf-v6-c-alert-group.pf-m-toast`.
5. Navigate back to the **Progress** tab and trigger a workspace error (e.g.
   network interruption) so that an alert toast appears.
6. Dismiss the alert and immediately re-run axe — verify no empty list violation
   appears.
7. Confirm that toast alerts still appear in the top-right corner of the
   viewport with the correct PatternFly styling.

### DIFF
--- a/packages/dashboard-frontend/src/components/AppAlertGroup/index.tsx
+++ b/packages/dashboard-frontend/src/components/AppAlertGroup/index.tsx
@@ -12,6 +12,7 @@
 
 import { Alert, AlertActionCloseButton, AlertGroup, AlertVariant } from '@patternfly/react-core';
 import React from 'react';
+import ReactDOM from 'react-dom';
 
 import { lazyInject } from '@/inversify.config';
 import { AppAlerts } from '@/services/alerts/appAlerts';
@@ -96,7 +97,14 @@ class AppAlertGroup extends React.PureComponent<Props, State> {
     if (alerts.length === 0) {
       return <></>;
     }
-    return <AlertGroup isToast>{alerts.map(alert => this.getAlert(alert))}</AlertGroup>;
+    // Use ReactDOM.createPortal directly (synchronous commit-phase cleanup) instead of
+    // AlertGroup isToast (which uses useEffect for portal cleanup, leaving an empty
+    // <ul role="list"> briefly visible between React commit and passive-effect teardown).
+    // The empty list violates WCAG 4.1.2 / aria-required-children (CRW-10740).
+    return ReactDOM.createPortal(
+      <AlertGroup className="pf-m-toast">{alerts.map(alert => this.getAlert(alert))}</AlertGroup>,
+      document.body,
+    );
   }
 }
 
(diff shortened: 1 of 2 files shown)


dmytro-ndp (approved):
@olexii4 : I tested Eclipse Che Next with User Dashboard quay.io/eclipse/che-dashboard:pr-1572 on the board.

There was no empty UL element with class `pf-v6-c-alert-group pf-m-toast` by default, when error alert was absent at start workspace page.

When User Dashboard displayed UL element with class `pf-v6-c-alert-group pf-m-toast` and role="list", it included `li` element, as expected:

Image

Well done!

svor (approved):

PR #1557, opened 2026-05-06
fix(a11y): add session timeout warning modal (WCAG 2.2.1)

olexii4:
### What does this PR do?

Adds a session timeout warning modal to satisfy [WCAG 2.2.1 Timing Adjustable (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/timing-adjustable.html).

When a user sits idle on any Dashboard page, the OAuth proxy session eventually expires without any prior warning, causing an abrupt redirect to the login page and potential loss of unsaved form data.

This PR:

- **Backend** — exposes `sessionTimeout` (seconds) through the existing `/dashboard/api/server-config` endpoint using a two-source strategy:
  1. **Primary**: reads `spec.networking.auth.gateway.oAuthProxy.cookieExpireSeconds` from the `CheCluster` CR ([added in che-operator#1760](https://github.com/eclipse-che/che-operator/pull/1760), `+kubebuilder:default:=86400`).
  2. **Fallback**: when the CR field is absent (older operators), parses `cookie_expire` directly from the `che-gateway-config-oauth-proxy` ConfigMap — which always reflects the value the OAuth proxy actually enforces.
  3. Returns `-1` when neither source is available (e.g. local dev without k8s) so the modal is a no-op.

- **Frontend** — adds a `useSessionTimeout` hook that tracks UI inactivity, and a `SessionTimeoutModal` component (mounted globally in `App.tsx`) that:
  - appears 60 seconds before the session cookie expires
  - shows a live countdown ring (amber → red/pulsing at ≤ 20 s, suppressed by `prefers-reduced-motion`)
  - offers **Extend session** (primary button) and **Sign out now** (link button)
  - resets the OAuth proxy session cookie by calling `GET /dashboard/api/user/id` on extend
  - signs out automatically when the countdown reaches zero
  - ignores mouse movement while visible — only explicit button/keyboard interaction dismisses it

The modal is a no-op when `sessionTimeout ≤ 0`.

### Screenshot/screencast of this PR

**Normal state (> 20 s remaining) — amber ring:**
Знімок екрана 2026-05-07 о 14 21 03

**Urgent state (≤ 20 s remaining) — red pulsing ring:**
Знімок екрана 2026-05-07 о 04 54 17

### What issues does this PR fix or reference?

fixes https://issues.redhat.com/browse/CRW-10290

### Is it tested? How?

Tested manually on a local CRC cluster (OpenShift + HTPasswd OAuth) and a Minikube cluster (Dex OIDC).

[...]

PR #1576, opened 2026-05-19
Fix session timeout warning

olexii4:
# What does this PR do?

Fixes several bugs in the session-timeout warning modal (#1557) and adds silent
auto-extend for active users so they never experience unexpected sign-outs.

---

### Bug 1 — Wrong selector fallback

`selectSessionTimeout` used `?? 86400` — if the backend omits `sessionTimeout`,
the modal activated with a 24 h timeout instead of staying disabled.
Changed to `?? -1`.

### Bug 2 — JS idle timer vs OAuth cookie mismatch

The hook reset the idle timer on UI events (`mousemove`, `keydown`, …) but the
OAuth cookie is refreshed by **HTTP requests through the gateway** (including
background workspace polling). Added an Axios response interceptor that calls
`startIdleTimer()` on every successful response, keeping both clocks in sync.

### Bug 3 — No guard for very short sessions

Added `MIN_SESSION_TIMEOUT_SECONDS = 90`. The modal is suppressed for sessions
shorter than 90 s (< 30 s of warning time — not useful).

### Bug 4 — Broken keep-alive URL

`KEEP_ALIVE_URL` was set to  `/dashboard/api/server-config`, which always returns 200, is a
lightweight read, and passes through the OAuth proxy (which refreshes the session
cookie on the request).### Screenshot/screencast of this PR

### Is it tested? How?

 2. Apply the next patch
```bash
kubectl patch checluster/eclipse-che -n eclipse-che --type=merge \
  -p '{"spec":{"networking":{"auth":{"gateway":{"oAuthProxy":{"cookieExpireSeconds":120}}}}}}'
```

3. **Idle path**: open dashboard, do not move the mouse.
  Modal appears at 60 s; sign-out at 120 s.
4. **Active path**: keep moving the mouse.
  No modal; session silently extended every time idle exceeds 45 s.


svor:
@SkorikSergey please take a look

SkorikSergey (approved):
Looks good to merge.

svor (approved):
lgtm

PR #1578, opened 2026-05-21
fix: preserve docker config when creating secret on update fallback

vinokurig:
### What does this PR do?
When updating a docker config secret fails with 404, the service falls back to creating a new secret. Previously, the `dockerCfg` parameter was not passed to `createNamespacedSecret`, causing it to create an empty config instead of preserving the provided configuration.

https://redhat.atlassian.net/browse/CRW-10854

### Is it tested? How?

1. Go to `User Preferences` -> `Container Registries` tab and click the `Add Container Registry` button, see: new registry item is created. 
2. Switch to another tab and then go back to the `Container Registries` tab, see: the registry item is still present in the list.


olexii4 (approved):
LGTM
Good catch on the missing argument in the fallback path.

PR #1579, opened 2026-05-21
fix: swagger route and local run improvements

### LINKED ISSUE #23894: [UD] Swagger UI route served under `/dashboard/api/swagger` instead of `/dashboard/swagger`; local development OAuth and stability gaps
## Describe the bug

Three independent problems are bundled here because they all block the same local-development workflow.

---

### 1. Swagger route prefix conflicts with API namespace

The Swagger UI and its JSON spec are mounted at `/dashboard/api/swagger`, which places documentation inside the `/dashboard/api/` namespace reserved for backend API endpoints. This causes two problems:

- Any route guard, proxy rule, or middleware that blanket-applies to `/dashboard/api/**` (authentication, CORS, rate-limiting) also applies to the documentation, making the spec unreachable in secured deployments.
- The OpenAPI JSON URL embedded in the Swagger UI resolves incorrectly when accessed through the Che gateway, returning a 404.

**Expected route:** `/dashboard/swagger` (documentation separate from API)
**Actual route:** `/dashboard/api/swagger`

---

### 2. Local run on OpenShift uses the dashboard service account token instead of the user's OAuth token

When running locally against a CRC/OpenShift cluster, `local-run.sh` forwards the dashboard service account token as `CLUSTER_ACCESS_TOKEN`. This is the wrong identity: the service account token belongs to the `che-dashboard` pod, not to the developer. Workspace API calls that require the user's own RBAC permissions (e.g. creating a DevWorkspace in the user's namespace) succeed only by accident — because the service account happens to have broad cluster rights — and break on tighter clusters where it does not.

The correct token is the developer's own OAuth token, obtainable via `oc whoami -t`. That token is already present on the developer's machine but was never wired into the local-run flow. As a result:

- Workspace operations run under the service account identity, not the user's, masking permission errors that would appear in production.
- On clusters where `oc` is not available or the developer is not logged in, the service account token is unavailable too, so all workspace API calls fail with 401 and no clear error is surfaced.
- The Minikube/dex code path authenticates the developer through a browser login window; OpenShift has no equivalent, creating inconsistent UX between the two local-run modes.

---

### 3. Local-run pod resolution matches `Terminating` pods during rollouts

`local-run.sh` resolves the dashboard pod by label selector without filtering on phase. During any image rollout there are two pods — one `Running`, one `Terminating` — and the script can pick the terminating one, causing the port-forward to fail immediately.

The dex configMap patch also uses `replace` mode; if the Che operator detects the change and overwrites it, the local configMap diverges silently and OAuth breaks.

---

### Swagger route (issue 1)

1. Deploy Eclipse Che.
2. Open `{che-url}/dashboard/api/swagger` — the Swagger UI loads.
3. Inspect the "Servers" dropdown in the Swagger UI — the base URL shown includes `/api/swagger`.
4. From a terminal: `curl {che-url}/dashboard/api/swagger/json` — observe it returns the spec.
5. Open `{che-url}/dashboard/swagger` — 404.

### Local run on OpenShift (issue 2)

1. Clone `che-dashboard`, install dependencies.
2. Connect `oc` to a CRC cluster.
3. Run `yarn start -c` (local-run mode against OpenShift inside Eclipse-che).
4. Open `http://localhost:8080/dashboard/` — dashboard loads without any browser login prompt.
5. Try to create or start a workspace — observe the request succeeds only as long as the `oc whoami -t` token is valid; rotate the token and the session breaks with no recovery path.

### Terminating pod resolution (issue 3)

1. Start `yarn start` with a running port-forward.
2. Trigger a pod rollout (e.g., patch the dashboard image).
3. Observe the port-forward drops and the script does not reconnect to the new `Running` pod.

---

## Expected behavior

1. Swagger UI and spec are served at `/dashboard/swagger` and `/dashboard/swagger/json`. The `/dashboard/api/` namespace contains only backend API endpoints.
2. When running locally against an OpenShift cluster, the dashboard opens a browser OAuth window (same UX as Minikube/dex) and exchanges a short-lived token. `oc whoami -t` is used only as a fallback for `CLUSTER_ACCESS_TOKEN` when the OAuth flow is unavailable (`yarn start -c`).
3. Pod resolution in `local-run.sh` filters to `Running` phase only, ensuring port-forward always targets the live pod.

---

## Che version

next (development version)

## Runtime

OpenShift (CRC for local development)

## Installation method

chectl/next; local dev with `yarn start`

## Environment

macOS

olexii4:
### What does this PR do?

**fix(swagger):** Changes the Swagger UI/JSON route prefix from `/dashboard/api/swagger` to `/dashboard/swagger`, keeping documentation endpoints separate from API endpoints.

**feat(run): `local-patch.sh`** — new script that builds the dashboard image for the local CRC VM architecture, loads it directly into the VM via SSH (bypassing any container registry), and patches the CheCluster CR to use it.

**fix(run): local run stability**
- Filter only `Running` pods when resolving the dashboard pod name — avoids matching `Terminating` pods during rollouts.
- Suppress `oc whoami -t` stderr on non-OpenShift clusters.
- Change dex configMap patch from replace to append so the operator doesn't detect a breaking change.
- Remove the now-unnecessary `che-gateway-config-oauth-proxy` patching block.

**fix(local-run): OpenShift browser OAuth flow** — when running locally against a CRC/OpenShift cluster, the dashboard now shows a browser login window (same UX as Minikube/dex) instead of silently using `oc whoami -t`.
- `prepare-local-run.sh` reads the OAuthClient name from the `che-gateway-config-oauth-proxy` ConfigMap, patches the OAuthClient with `http://localhost:8080/oauth/callback` as a redirect URI, and sets `grantMethod=auto` to skip the grant confirmation screen.
- `local-run.sh` detects the oauth-proxy gateway first, exports `OPENSHIFT_OAUTH_URL` / `CLIENT_ID` / `CLIENT_SECRET`, and sets `NODE_EXTRA_CA_CERTS` so the backend's token exchange can verify the CRC TLS certificate.
- `revert-local-run.sh` removes the localhost redirect URI and restores `grantMethod=prompt` on cleanup.
- New `openshiftOauth2.ts` plugin reuses `@fastify/oauth2` pointed at `/oauth/authorize` and `/oauth/token` on the OpenShift OAuth server; existing `dexCallback`, `signOut`, and `authorizationHooks` are reused unchanged.
- che-in-che mode (`yarn start -c`) skips the OAuth flow and keeps the direct token path.

**fix(local-run): `CLUSTER_ACCESS_TOKEN` from `oc whoami -t`** — on OpenShift local run, `getServiceAccountToken` now prefers `CLUSTER_ACCESS_TOKEN` (the developer's own OAuth token) over the dashboard service account token so workspace operations run under the correct identity.

**fix(bootstrap): skip internal registry fetches when `disableInternalRegistry: true`** — the bootstrap unconditionally fetched the internal devfile registry before reading the flag, causing the await chain to abort so external registries were never loaded and the samples list stayed empty. Internal fetches are now gated on `!isInternalRegistryDisabled`; external registries always load.

**fix(a11y): PF6 `selectableActions` pattern for clickable cards** — two components were emitting PatternFly v6 console errors:
- `SampleCard`: replaced ad-hoc `onClick`/`role="button"` with `selectableActions.onClickAction` + `selectableActionAriaLabel`; tags moved from `CardHeader actions` (prohibited on clickable-only cards) into `CardBody`.
- `EditorSelectorEntry`: replaced `selectableActionAriaLabelledby` (indirect DOM reference) with explicit `selectableActionAriaLabel`.

**fix(watch): null guard in `notifyWatchError`** — when a K8s watch stream closes normally (e.g. port-forward reset), the done callback receives `null`; accessing `.statusCode` on it crashed the error handler in `podApi`, `eventApi`, and `devWorkspaceApi`.

**fix(local-run): remove devfile registry proxy** — a proxy for `/dashboard/devfile-registry/` forwarded requests to the upstream cluster, but the che-gateway returned 302 (auth redirect). The local `lib/public/dashboard/devfile-registry` directory is already populated by `local-run.sh`, so the static file server handles it correctly without the proxy.

fixed https://github.com/eclipse-che/che/issues/23894

### Is it tested? How?

- Swagger UI accessible at `/dashboard/swagger`; `/dashboard/api/swagger` returns 404.
- External devfile registry samples with local-run visible when `disableInternalRegistry: true` + `externalDevfileRegistries` set.
- No PF6 Card console warnings on GetStarted (sample cards) or Create Workspace (editor selector).
- OpenShift browser OAuth flow: `yarn start:prepare && yarn start` on CRC → browser redirects to OpenShift login → after credentials → redirects back to dashboard — verified on CRC.
- che-in-che mode (`yarn start -c`) unaffected — continues to use `oc whoami -t` token.
- Watch error (`Cannot read properties of null (reading 'statusCode')`) no longer appears in local run logs.


tolusha:

**Available commands**:

olexii4:
/che-ai-assistant ok-pr-review

Review is complete. Please check the review comments below.

tolusha (commented):
## Comprehensive PR Review

I've completed a multi-layered review (standard, deep, and impact analysis) of this PR. Here's a summary of findings:

### 🔴 Blocking Issue

**Missing SampleCard component changes**: The PR description states that `SampleCard` was refactored to use `selectableActions.onClickAction` + `selectableActionAriaLabel` with tags moved from `CardHeader` to `CardBody`, but only the CSS file (`index.module.css`) appears in the diff. The component file (`index.tsx`) is missing. The `.sampleCardTags` CSS class is currently unused. Please clarify if these changes were dropped during rebase or if the PR description needs updating.

---

### ⚠️ Important Warnings

1. **Race condition in `revert-local-run.sh`** (lines 63-67): The script removes the localhost redirect URI using array index-based removal. Between the `jq` query and `oc patch`, the Che operator could modify the array, causing removal of the wrong URI. Consider using a value-based filter instead of index-based removal.

2. **Redundant check in bootstrap** (`bootstrap/index.ts` ~line 648): The outer guard checks `if (!isInternalRegistryDisabled)`, but inside there's another `if (devfileRegistry?.disableInternalRegistry !== true)` which is always true within the outer block. This is dead code.

3. **Backward compatibility** (`swagger.ts`): The route change from `/dashboard/api/swagger` to `/dashboard/swagger` is undocumented as a breaking change. Consider adding a 301 redirect or documenting in release notes.

4. **Missing test coverage**:
   - `openshiftOauth2.ts` has no unit tests (auth-adjacent code)
   - Bootstrap flag behavior (`disableInternalRegistry=true` path) has no test

---

### 💡 Suggestions

1. **Fragile ConfigMap parsing**: The `awk -F'"' '/client_id/{print }'` pattern (repeated 3x) assumes double-quoted values. Consider documenting the format or switching to `yq`/`jq`.

2. **Hardcoded localhost** (`openshiftOauth2.ts:21`): `CHE_HOST = 'http://localhost:8080'` is hardcoded. Consider making this configurable for future flexibility.

3. **`check-resolutions.js`** depends on `semver` without it being a direct dependency. It works as a transitive dep but could break.

4. **SSH host key verification disabled** (`local-patch.sh:43`): While reasonable for CRC VM, consider adding a comment explaining why this is safe.

---

### ✅ Positive Feedback

- **Excellent PR description**: Each fix is clearly explained with context
- **Clean OAuth implementation**: Reuses `@fastify/oauth2` and existing auth infrastructure
- **Proper lifecycle**: prepare→run→revert with idempotent prepare
- **Consistent defensive coding**: Null guard pattern applied uniformly
- **Well-structured tooling**: `local-patch.sh` has good help text and error handling
- **Thorough test coverage** for `registryApi.external.spec.ts` and `getServiceAccountToken.spec.ts`

---

### 📊 Review Summary

**Does This PR Solve the Issue?**
- ✅ Issue #1 (Swagger route): Fixed
- ✅ Issue #2 (Local-run OAuth): Implemented
- ✅ Issue #3 (Pod resolution): Fixed

**Design Quality**: Sound overall - clean abstraction reuse, proper lifecycle management

**Evolvability**: Good for current scope. Adding a 3rd auth provider would touch 5 files (borderline). Hardcoded values like `CHE_HOST` reduce flexibility.

**Operational Readiness**: No major concerns. Good error handling, secret handling, and resource cleanup.

---

### 🎯 Verdict

**Request Changes** - Please address the blocking issue (#1) and consider the important warnings (#2-4) before merge.

---

*Review conducted with: /ok-pr-review:summary → /ok-pr-review:review → /ok-pr-review:deep-review → /ok-pr-review:impact*

dmytro-ndp:
/che-ai-assistant check-pr-test-failures

Review is complete. Please check the review comments below.

olexii4:
/che-ai-assistant check-pr-test-failures

Review is complete. Please check the review comments below.

akurinnoy (commented):
The PR overall looks good, couple questions and suggestions below.

packages/dashboard-frontend/src/services/bootstrap/index.ts
```diff
@@ -340,41 +340,42 @@ export default class Bootstrap {
...
-    if (
-      devfileRegistry?.disableInternalRegistry !== undefined &&
-      devfileRegistry?.disableInternalRegistry !== true
-    ) {
-      const airGapedSampleURL = new URL('/dashboard/api/airgap-sample', window.location.origin)
-        .href;
-      await requestRegistriesMetadata(airGapedSampleURL, false)(
+    const isInternalRegistryDisabled = devfileRegistry?.disableInternalRegistry === true;
+
+    if (!isInternalRegistryDisabled) {
```
akurinnoy, line 347: The refactor gates `DEFAULT_REGISTRY` and `gettingStartedSampleURL` fetches behind `!isInternalRegistryDisabled`. Previously both were fetched unconditionally - only the airgap-sample endpoint was gated on `disableInternalRegistry`.

This means when `disableInternalRegistry === true`, the GetStarted page may now show no samples at all (only external registries, if configured). Is this the intended behavior? If so, would be good to call it out explicitly in the PR description since it is a user-visible change.

packages/dashboard-frontend/src/services/bootstrap/index.ts
```diff
@@ -340,41 +340,42 @@ export default class Bootstrap {
...
+        '/dashboard/api/getting-started-sample',
+        window.location.origin,
+      ).href;
+      await requestRegistriesMetadata(gettingStartedSampleURL, false)(
+        this.store.dispatch,
+        this.store.getState,
+        undefined,
+      );
+
+      if (devfileRegistry?.disableInternalRegistry !== true) {
```
akurinnoy: This inner guard is always true - we are already inside `if (!isInternalRegistryDisabled)` where `isInternalRegistryDisabled = devfileRegistry?.disableInternalRegistry === true`. The `if` wrapper can be removed, leaving just the airgap fetch:

```suggestion
      const airGapedSampleURL = new URL('/dashboard/api/airgap-sample', window.location.origin)
        .href;
      await requestRegistriesMetadata(airGapedSampleURL, false)(
        this.store.dispatch,
        this.store.getState,
        undefined,
      );
```

packages/dashboard-frontend/src/pages/GetStarted/SamplesList/Gallery/Card/index.module.css
```diff
@@ -36,3 +36,10 @@ html.pf-v6-theme-dark .sampleCardIcon {
   background-color: var(--pf-t--color--gray--10);
   border-radius: calc(var(--pf-v6-c-card--BorderRadius) / 4);
 }
+
+.sampleCardTags {
```
akurinnoy: The `.sampleCardTags` CSS class is added here, but the `SampleCard` component (`index.tsx`) that would use it is not in the diff. The PR description says SampleCard was refactored (`selectableActions.onClickAction`, tags moved to `CardBody`) - were those changes lost during a rebase, or does the PR description need updating?

packages/dashboard-backend/src/localRun/plugins/openshiftOauth2.ts
```diff
@@ -0,0 +1,53 @@
...
+import oauth2Plugin, { OAuth2Namespace } from '@fastify/oauth2';
+import { FastifyInstance } from 'fastify';
+
+declare module 'fastify' {
+  interface FastifyInstance {
+    localStart: OAuth2Namespace;
+  }
+}
+
+const CHE_HOST = 'http://localhost:8080';
```
akurinnoy: `CHE_HOST` is hardcoded here and replicated independently in `prepare-local-run.sh` and `revert-local-run.sh` - three places that must stay in sync. If a developer runs the backend on a different port, they need to edit TypeScript source and re-compile.

Would it make sense to read from an env var with a fallback?

```suggestion
const CHE_HOST = process.env.CHE_HOST || 'http://localhost:8080';
```

run/local-patch.sh
```diff
@@ -0,0 +1,149 @@
...
+  sed -n '2,/^$/{ s/^# \?//; p }' "$0"
+  exit 0
+fi
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_DIR="$(dirname "$SCRIPT_DIR")"
+
+IMAGE_TAG="local-$(date +%Y%m%d-%H%M%S)"
+CHE_DASHBOARD_IMAGE="${CHE_DASHBOARD_IMAGE:-quay.io/local/che-dashboard:${IMAGE_TAG}}"
+BUILD_PLATFORM="${BUILD_PLATFORM:-linux/arm64}"
```
akurinnoy: `BUILD_PLATFORM` defaults to `linux/arm64`, which is correct for Apple Silicon but will silently build the wrong architecture on x86 hosts. Consider auto-detecting, e.g.:

```bash
DEFAULT_PLATFORM="linux/$(uname -m | sed 's/x86_64/amd64/' | sed 's/aarch64/arm64/')"
BUILD_PLATFORM="${BUILD_PLATFORM:-$DEFAULT_PLATFORM}"
```

run/local-run.sh
```diff
@@ -108,19 +108,58 @@ if [ ! -d $DASHBOARD_FRONTEND/lib/public/dashboard/devfile-registry ]; then
...
+GATEWAY_CHECK=$(kubectl get deployments.apps -n $CHE_NAMESPACE che-gateway --ignore-not-found -o=json | jq -e '.spec.template.spec.containers|any(.name == "oauth-proxy")')
+if [[ "$GATEWAY_CHECK" == "true" ]] && [[ -n "$(oc whoami -t 2>/dev/null)" ]] && [[ "$CHE_IN_CHE" == "false" ]]; then
+  # OpenShift native auth: use browser-based OAuth instead of oc whoami -t token
+  # (skipped in che-in-che mode where the user token is injected directly)
+  echo '[INFO] Detected OpenShift native auth mode. Setting up browser OAuth flow.'
+
+  # Read CLIENT_ID from the oauth-proxy ConfigMap (config is file-based, not CLI args)
+  export CLIENT_ID=$(kubectl get configmap che-gateway-config-oauth-proxy -n "$CHE_NAMESPACE" \
+    -o jsonpath='{.data.oauth-proxy\.cfg}' 2>/dev/null | \
+    awk -F'"' '/client_id/{print $2}')
```
akurinnoy, line 120: The `awk -F'"' '/client_id/{print $2}'` pattern for extracting `client_id` from the oauth-proxy ConfigMap is duplicated in `local-run.sh`, `prepare-local-run.sh`, and `revert-local-run.sh`. If the ConfigMap format ever changes, all three copies need updating independently. Consider extracting to a shared helper.

Non-blocking - just a maintenance thought for later.

packages/dashboard-frontend/src/services/bootstrap/index.ts
olexii4: I have fixed it.

packages/dashboard-frontend/src/services/bootstrap/index.ts
olexii4: I have fixed it.

packages/dashboard-backend/src/localRun/plugins/openshiftOauth2.ts
olexii4: I have fixed it.

run/local-run.sh
olexii4: I have fixed it.

run/local-patch.sh
olexii4: I have fixed it.

packages/dashboard-frontend/src/pages/GetStarted/SamplesList/Gallery/Card/index.module.css
olexii4: I have fixed it.

olexii4:
/che-ai-assistant generate-che-doc

Created documentation PR: https://github.com/eclipse-che/che-docs/pull/3158

run/local-run.sh
olexii4: I have fixed.

run/local-patch.sh
olexii4: I have fixed.

packages/dashboard-backend/src/localRun/plugins/openshiftOauth2.ts
olexii4: I have fixed.

packages/dashboard-frontend/src/pages/GetStarted/SamplesList/Gallery/Card/index.module.css
olexii4: I have fixed.

packages/dashboard-frontend/src/services/bootstrap/index.ts
olexii4: I have fixed.

packages/dashboard-frontend/src/services/bootstrap/index.ts
olexii4: I have fixed.

olexii4:
/che-ai-assistant ok-pr-review

Task completed.

run/local-patch.sh
```diff
@@ -0,0 +1,150 @@
...
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_DIR="$(dirname "$SCRIPT_DIR")"
+
+IMAGE_TAG="local-$(date +%Y%m%d-%H%M%S)"
+CHE_DASHBOARD_IMAGE="${CHE_DASHBOARD_IMAGE:-quay.io/local/che-dashboard:${IMAGE_TAG}}"
+DEFAULT_PLATFORM="linux/$(uname -m | sed 's/x86_64/amd64/' | sed 's/aarch64/arm64/')"
+BUILD_PLATFORM="${BUILD_PLATFORM:-$DEFAULT_PLATFORM}"
+CRC_SSH_PORT="${CRC_SSH_PORT:-2222}"
+TAR_PATH="/tmp/che-dashboard.tar"
```
akurinnoy, line 46: `TAR_PATH` is removed only on the happy path (`rm -f` after the successful scp). If `scp` or any step between `docker save` and the final `rm` fails, the local tar file stays on disk. A large dashboard image can be several hundred MB.

Adding a trap handles all exit paths:

```suggestion
TAR_PATH="/tmp/che-dashboard.tar"

trap 'rm -f "${TAR_PATH}"' EXIT
```

run/local-patch.sh
```diff
@@ -0,0 +1,150 @@
...
+
+CHECLUSTER_CR_NAME=$(kubectl get checluster -n "$CHE_NAMESPACE" -o jsonpath='{.items[0].metadata.name}')
+PREVIOUS_IMAGE=$(kubectl get checluster -n "$CHE_NAMESPACE" "$CHECLUSTER_CR_NAME" -o=json | jq -r '.spec.components.dashboard.deployment.containers[0].image')
+
+PATCH_VALUE="{\"deployment\": {\"containers\": [{\"image\": \"${CHE_DASHBOARD_IMAGE}\", \"name\": \"che-dashboard\", \"imagePullPolicy\": \"IfNotPresent\"}]}}"
+
+if [ "$PREVIOUS_IMAGE" == "null" ]; then
+  kubectl patch -n "$CHE_NAMESPACE" "checluster/$CHECLUSTER_CR_NAME" --type=json \
+    -p="[{\"op\": \"replace\", \"path\": \"/spec/components/dashboard\", \"value\": ${PATCH_VALUE}}]"
+else
```
akurinnoy, line 143: When `PREVIOUS_IMAGE` is `null`, this replaces `/spec/components/dashboard` entirely with the new deployment block. This silently drops any other fields that were set there - `headerMessage`, custom env vars, resource limits, etc.

The update path (else branch) only touches `containers/0/image` and `containers/0/imagePullPolicy`, preserving everything else. Does it make sense to use a merge-patch approach for both cases so existing fields are always preserved?

```suggestion
if [ "$PREVIOUS_IMAGE" == "null" ]; then
  kubectl patch -n "$CHE_NAMESPACE" "checluster/$CHECLUSTER_CR_NAME" --type=merge \
    -p="{\"spec\":{\"components\":{\"dashboard\":{\"deployment\":{\"containers\":[{\"name\":\"che-dashboard\",\"image\":\"${CHE_DASHBOARD_IMAGE}\",\"imagePullPolicy\":\"IfNotPresent\"}]}}}}}"
else
```

packages/dashboard-backend/src/localRun/plugins/openshiftOauth2.ts
```diff
@@ -0,0 +1,53 @@
...
+      auth: {
+        authorizeHost: openShiftOAuthUrl,
+        authorizePath: '/oauth/authorize',
+        tokenHost: openShiftOAuthUrl,
+        tokenPath: '/oauth/token',
+      },
+    },
+    scope: ['user:full'],
+    startRedirectPath: '/oauth/sign_in',
+    callbackUri: `${CHE_HOST}/oauth/callback`,
```
akurinnoy, line 51: `server.register(oauth2Plugin, {...})` does not set `httpOptions.timeout`. The underlying `simple-oauth2` library defaults to no timeout. If the OpenShift OAuth server is unreachable (CRC wake from sleep, VPN change), the token exchange POST hangs indefinitely with no user-visible error.

Consider adding a timeout:

```suggestion
  server.register(oauth2Plugin, {
    name: 'localStart',
    credentials: {
      client: {
        id: clientId,
        secret: clientSecret,
      },
      auth: {
        authorizeHost: openShiftOAuthUrl,
        authorizePath: '/oauth/authorize',
        tokenHost: openShiftOAuthUrl,
        tokenPath: '/oauth/token',
      },
    },
    httpOptions: {
      timeout: 10000,
    },
    scope: ['user:full'],
    startRedirectPath: '/oauth/sign_in',
    callbackUri: `${CHE_HOST}/oauth/callback`,
  });
```

packages/dashboard-backend/src/devworkspaceClient/services/__tests__/registryApi.external.spec.ts
```diff
@@ -44,21 +46,25 @@ jest.mock('@/devworkspaceClient/services/helpers/prepareCoreV1API', () => ({
...
+    const kubeConfig = new KubeConfig();
+    kubeConfig.makeApiClient = jest
+      .fn()
+      .mockImplementation(() => mockCustomObjectsApi as unknown as CustomObjectsApi);
+
+    const saKubeConfig = new KubeConfig();
+    saKubeConfig.makeApiClient = jest
+      .fn()
+      .mockImplementation(() => mockSaCustomObjectsApi as unknown as CustomObjectsApi);
+
```
akurinnoy, line 68: `clearAllMocks()` fires after `new RegistryApiService(...)`. This wipes any mock call counts that happened during construction. If a future test checks how many times `makeApiClient` was called (including the constructor call), it will always see zero.

Moving `clearAllMocks()` before the mock setup is the safer pattern:

```suggestion
    jest.clearAllMocks();
```

...at the top of `beforeEach`, before the `KubeConfig` setup.

run/local-run.sh
```diff
@@ -108,19 +108,58 @@ if [ ! -d $DASHBOARD_FRONTEND/lib/public/dashboard/devfile-registry ]; then
...
+      echo '[INFO] Evaluated Dex ingress. Looking for staticClientID and staticClientSecret...'
+      export CLIENT_ID=$(kubectl get -n dex configMaps/dex -o jsonpath="{.data['config\.yaml']}" | yq ${YQ_FLAGS} ".staticClients[0].id" -)
+      export CLIENT_SECRET=$(kubectl get -n dex configMaps/dex -o jsonpath="{.data['config\.yaml']}" | yq ${YQ_FLAGS} ".staticClients[0].secret" -)
+      echo '[INFO] Done. $CLIENT_ID and $CLIENT_SECRET are set.'
+    fi
   fi
 fi
 
-DASHBOARD_POD_NAME=$(kubectl get pods -n "$CHE_NAMESPACE" -o=custom-columns=:metadata.name | grep dashboard)
+DASHBOARD_POD_NAME=$(kubectl get pods -n "$CHE_NAMESPACE" --field-selector=status.phase=Running -o=custom-columns=:metadata.name | grep dashboard | head -1)
```
akurinnoy, line 162: When no Running dashboard pod exists (during a rollout or after a crash), `DASHBOARD_POD_NAME` is empty. The subsequent `kubectl exec "$DASHBOARD_POD_NAME"` fails with a confusing error.

An explicit guard would produce a clearer message:

```suggestion
DASHBOARD_POD_NAME=$(kubectl get pods -n "$CHE_NAMESPACE" --field-selector=status.phase=Running -o=custom-columns=:metadata.name | grep dashboard | head -1)

if [[ -z "$DASHBOARD_POD_NAME" ]]; then
  echo "[ERROR] No running dashboard pod found in namespace $CHE_NAMESPACE"
  exit 1
fi
```

(Same issue is in `run/prepare-local-run.sh:41`.)

scripts/check-resolutions.js
```diff
@@ -0,0 +1,209 @@
...
+//   - all ranges that request the package (from headers + dependency blocks)
+//   - any "older" ranges that the resolution is currently overriding
+//
+// Usage: node scripts/check-resolutions.js
+//   or:  yarn resolutions:check
+
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
```
akurinnoy, line 25: `require('semver')` works because `semver` is pulled in transitively, but it is not listed in the root `package.json` `devDependencies`. If any dependency that currently pulls it in is updated or removed, this script breaks at runtime without a compilation error to catch it.

Consider adding `semver` as a direct devDependency.

akurinnoy (approved):
Nice grab bag - the OAuth local-run flow and the bootstrap fix for external registries are solid additions.

Two general observations:

**Missing test for `getServiceAccountToken` exit path** - The new tests cover `CLUSTER_ACCESS_TOKEN` priority and `SERVICE_ACCOUNT_TOKEN` fallback, but not the `process.exit(1)` branch (when neither token is set in local-run mode). This is the only branch with a side effect that kills the process.

**`check-resolutions.js` does not parse `optionalDependencies` blocks** - The yarn.lock parser handles `dependencies:` and `peerDependencies:` blocks but skips `optionalDependencies:`. Ranges in optional dependency blocks are not collected, so the script can report a resolution as "possibly removable" when it is still overriding a range that only appears in an `optionalDependencies:` block. Not a blocker - just something to be aware of if the script reports false positives.

scripts/check-resolutions.js
olexii4: I have fixed it.

run/local-run.sh
olexii4: I have fixed it.

run/local-patch.sh
olexii4: I have fixed it.

run/local-patch.sh
olexii4: I have fixed it.

packages/dashboard-backend/src/localRun/plugins/openshiftOauth2.ts
olexii4: I have fixed it.

packages/dashboard-backend/src/devworkspaceClient/services/__tests__/registryApi.external.spec.ts
olexii4: I have fixed it.

PR #1569, opened 2026-05-18
refactor: remove backend user profile route and use Che server endpoint directly

vinokurig:
### What does this PR do?
In some circumstances, e.g., a high-load OpenShift cluster, creating or updating a Kubernetes secret may take some time (it is not instant). If reading the `user-profile` secret occurs during the create or update operation of the secret, an error is thrown. In order to avoid this race condition, request the user from the `che-server` API instead of reading the `user-profile` secret directly.

**DO NOT MERGE** until https://github.com/eclipse-che/che-server/pull/1018 is merged.
### What issues does this PR fix or reference?
https://redhat.atlassian.net/browse/CRW-10682

### Is it tested? How?

N/A

PR #1587, opened 2026-05-27
refactor: simplify user profile to username-only state

vinokurig:
### What does this PR do?
In some circumstances, e.g., a high-load OpenShift cluster, creating or updating a Kubernetes secret may take some time (it is not instant). If reading the `user-profile` secret occurs during the create or update operation of the secret, an error is thrown. In order to avoid this race condition, request the user from the `che-server` API instead of reading the `user-profile` secret directly.

Removed the full user profile API endpoint and state management in favor of a simpler username-only approach. The previous IUserProfile interface contained both email and username, but only username was actively used throughout the application.

Changes:
- Backend: removed /api/userprofile endpoint and userProfileApi service
- Frontend: renamed User/Profile store slice to User/Name and removed the redundant gravatar icon
- API: removed IUserProfile interface from common DTO
- Simplified API response to return username string directly
- Updated all imports and selectors to use new username-only state
- Renamed actionCreators from userProfileActionCreators to usernameActionCreators
- Updated all tests to reflect new naming and structure

https://redhat.atlassian.net/browse/CRW-10682

### Is it tested? How?

N/A


olexii4 (changes requested):
The idea is correct — switching away from reading the `user-profile` secret during concurrent writes makes sense. But there are a couple of things that need to be addressed.

The [`gravatar-url`](https://www.npmjs.com/package/gravatar-url) package generates image URLs by hashing an email address. Passing a username instead of an email will never match any real Gravatar account — every user will get the same generated fallback avatar. So removing `email` from the profile breaks the avatar in the header. The `Avatar` component needs to be redesigned with this in mind — either use initials, a generic user icon, or find another way to get the email.

Also, the old code had `{isUserAuthenticated && (...)}` guards around `UserMenu` and `Avatar`. Those are gone now, so both components always render — including during bootstrap when the username is still `'unknown'`. That needs to be handled somehow.

My suggestion: instead of merging this PR, let's look at https://github.com/eclipse-che/che-dashboard/pull/1569 and fix it there. The original approach of reading the `user-profile` secret is fine — the only problem is the race condition on high-load clusters. That can be handled by adding a short delay and a few retries before giving up. That way we keep the email, the Gravatar works, and we don't need `che-server#1018` to land first.

@vinokurig @akurinnoy what do you think?

vinokurig:
@olexii4 Originally we do not have email either, we generate it from the username by adding `@che` to the username, [see](https://github.com/eclipse-che/che-server/blob/dd873a597bb4f11ac9adf8806f39eccbefab479d/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/configurator/UserProfileConfigurator.java#L70). 
So in origin it does not work as well.

olexii4:
> @olexii4 Originally we do not have email either, we generate it from the username by adding `@che` to the username, [see](https://github.com/eclipse-che/che-server/blob/dd873a597bb4f11ac9adf8806f39eccbefab479d/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/configurator/UserProfileConfigurator.java#L70).
> So in origin it does not work as well.

@vinokurig thanks, that settles it. If the email was always `<username>@che`, the Gravatar has never resolved to a real account — it has always been the `retro` generated fallback. This PR changes nothing observable for the avatar.

That makes me think we should just remove the `Avatar` entirely from `Tools/index.tsx`. There is nothing meaningful behind it:

- `gravatarUrl(email)` with a synthetic `@che` address never matches a real account
- the component always renders the same generated placeholder
- keeping it means keeping the `gravatar-url` dependency and the `isUserAuthenticated` guard for no practical benefit

Concretely, remove from `packages/dashboard-frontend/src/Layout/Header/Tools/index.tsx`:

```diff
-import gravatarUrl from 'gravatar-url';
...
-  const imageUrl = email ? gravatarUrl(email, { default: 'retro' }) : '';
-  const isUserAuthenticated = !!email;
...
-          {isUserAuthenticated && (
-            <ToolbarItem>
-              <Avatar src={imageUrl} alt="Avatar image" size="md" />
-            </ToolbarItem>
-          )}
```

@ibuziuk @svor — thoughts on dropping the `Avatar` altogether?

svor:
IMO it makes sense to remove the avatar since it doesn't provide any functionality or useful information - it's just a default icon. It would also make the UI more consistent with the OpenShift console.

vinokurig:
@olexii4 @svor 
removed the gravatar widget:
image

svor (approved):
lgtm

olexii4 (approved):
LGTM

tolusha:

**Available commands**:

PR #1595, opened 2026-06-01
Enable che-dashboard build support on s390x architecture

swapnilsingh-ibm:
### What does this PR do?
This PR enables support for building the Che Dashboard on the s390x architecture.
It updates and refines architecture compatibility logic and ensures successful builds on IBM Z running RHEL 9.2 (s390x).

Screenshot 2026-05-29 at 1 31 44 PM
Screenshot 2026-05-29 at 1 31 58 PM
Screenshot 2026-05-29 at 1 34 11 PM
Screenshot 2026-05-29 at 1 34 39 PM
Screenshot 2026-05-29 at 1 34 54 PM

This is part of internal enablement for Che on s390x (IBM Z).

### Is it tested? How?

Tested end-to-end on native IBM Z hardware (RHEL 9.2, s390x) and a dedicated s390x OpenShift 4.21 cluster:

  1. Native image build — `podman build -f build/dockerfiles/rhel.Dockerfile` completed successfully on s390x
  2. Architecture verification — `podman inspect` confirms `OS/Arch: linux/s390x`
  3. Full unit test suite — yarn test ran natively inside a ubi9/nodejs-20 container on s390x: `2576 tests passed, 369 suites, 251 snapshots — all green`
  4. Live OpenShift deployment — image deployed to a production-like s390x OpenShift 4.21 cluster (3 masters + 3 workers, all s390x). Dashboard pod ran
  1/1 Running on worker node `qe2-devspaces-s390x-worker-0` with image `quay.io/swasingh/che-dashboard:s390x-test`. Dashboard UI was verified in browser — workspace creation page loaded 
  successfully.
  5. Image published — `quay.io/swasingh/che-dashboard:s390x-test` is publicly available for independent reviewer verification

#### Release Notes

The che-dashboard container image now ships as a multi-arch manifest that includes `linux/s390x` (IBM Z) in addition to `linux/amd64` and `linux/arm64`. Users deploying Eclipse Che on IBM Z infrastructure can now pull the standard image without requiring a separate build.

#### Docs PR

Documentation updates are included directly in this PR:
  - `run/README.md` — updated supported architectures and build examples
  - `run/MULTIARCH_BUILD.md` — updated overview and added s390x build instructions
  
  No separate docs repo PR is required as the changes are self-contained to the `run/` directory docs.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -145,8 +145,7 @@ jobs:
           S390X_VERSION="${{ needs['build-images'].outputs.s390x }}"
 
           if [[ -z "$AMD64_VERSION" || \
-               -z "$ARM64_VERSION" || \
-               -z "$S390X_VERSION" ]]; then
+               -z "$ARM64_VERSION" ]]; then
             echo "[!] The job 'build-images' fails on some of the architectures. Can't create complete manifest.";
             exit 1;
           fi
@@ -154,8 +153,9 @@ jobs:
           AMEND=""
           AMEND+=" --amend ${{ env.IMAGE }}:$AMD64_VERSION";
           AMEND+=" --amend ${{ env.IMAGE }}:$ARM64_VERSION";
-          AMEND+=" --amend ${{ env.IMAGE }}:$S390X_VERSION";
-
+          if [ -n "$S390X_VERSION" ]; then
+              AMEND+=" --amend ${{ env.IMAGE }}:$S390X_VERSION";
+          fi
           docker manifest create ${{ env.IMAGE }}:${{ github.event.inputs.version }} $AMEND
           docker manifest push ${{ env.IMAGE }}:${{ github.event.inputs.version }}
       -
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -86,10 +86,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        arch: [amd64,arm64]
+        arch: [amd64,arm64,s390x]
     outputs:
       amd64: ${{ steps.result.outputs.amd64 }}
       arm64: ${{ steps.result.outputs.arm64 }}
+      s390x: ${{ steps.result.outputs.s390x }}
     steps:
       -
         name: "Checkout Che Dashboard source code"
@@ -141,9 +142,11 @@ jobs:
         run: |
           AMD64_VERSION="${{ needs['build-images'].outputs.amd64 }}"
           ARM64_VERSION="${{ needs['build-images'].outputs.arm64 }}"
+          S390X_VERSION="${{ needs['build-images'].outputs.s390x }}"
 
           if [[ -z "$AMD64_VERSION" || \
-               -z "$ARM64_VERSION" ]]; then
+               -z "$ARM64_VERSION" || \
+               -z "$S390X_VERSION" ]]; then
             echo "[!] The job 'build-images' fails on some of the architectures. Can't create complete manifest.";
             exit 1;
           fi
@@ -151,6 +154,7 @@ jobs:
           AMEND=""
           AMEND+=" --amend ${{ env.IMAGE }}:$AMD64_VERSION";
           AMEND+=" --amend ${{ env.IMAGE }}:$ARM64_VERSION";
+          AMEND+=" --amend ${{ env.IMAGE }}:$S390X_VERSION";
 
           docker manifest create ${{ env.IMAGE }}:${{ github.event.inputs.version }} $AMEND
           docker manifest push ${{ env.IMAGE }}:${{ github.event.inputs.version }}
--- a/run/MULTIARCH_BUILD.md
+++ b/run/MULTIARCH_BUILD.md
@@ -4,6 +4,9 @@
 
 The `build-multiarch.sh` script builds Docker images for multiple architectures (AMD64, ARM64, and s390x) and pushes them to a container registry.
 
+**Note**:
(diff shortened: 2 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -126,6 +127,7 @@ jobs:
         run: echo "${{ matrix.arch }}=${{ github.event.inputs.version }}-${{ matrix.arch }}" >> $GITHUB_OUTPUT
 
   create-manifest:
+    if: always()
```
olexii4: @swapnilsingh-ibm  The `create-manifest` uses `if: always()`, which runs the job even when the workflow is cancelled. The inner validation already requires all three arch outputs to be present and exits 1 if any are missing, so `if: always()` doesn't help with partial build failures — it just adds an unnecessary risk of running on cancellation.

```suggestion
    if: ${{ !cancelled() }}
```

.github/workflows/release.yml
```diff
@@ -126,6 +127,7 @@ jobs:
         run: echo "${{ matrix.arch }}=${{ github.event.inputs.version }}-${{ matrix.arch }}" >> $GITHUB_OUTPUT
 
   create-manifest:
+    if: ${{ !cancelled() }}
```
akurinnoy: The nightly workflow uses `if: always()` because it creates a partial manifest with whatever architectures succeeded. The release workflow is different - it requires all three and exits 1 if any are missing. `!cancelled()` still runs the job on failure, so if `build-images` fails, `create-manifest` runs and then immediately fails on the empty-variable check instead of being skipped.

The outcome is the same, but the GitHub Actions UI shows two red jobs instead of one red + one skipped. Consider removing it to keep the release workflow consistent with its strict-validation pattern.

akurinnoy (commented):
We had s390x builds in the release workflow before, but they failed frequently and blocked releases for the other architectures. That is why s390x was dropped entirely.

Before we restore it, we need clarity on ownership: if s390x builds start failing again, who will be responsible for investigating and fixing them? Without a clear maintainer, we risk repeating the same cycle - adding the architecture, having it break, and then removing it again.

Could you clarify the maintenance plan for s390x going forward?

cc @ibuziuk @svor

.github/workflows/release.yml
olexii4: ok

swapnilsingh-ibm: Hi @akurinnoy, thank you for the context.

- Myself(@swapnilsingh-ibm) and my colleague Sudharshan(@sudharshanibm) from IBM will take the responsibility for any s390x-specific failures, or investigations.

- Additionally, for reference s390x support has already been merged across other Eclipse Che components as follows. This is part of an active IBM effort to enable the full Che ecosystem on IBM Z, not a one-off addition.
eclipse-che/che-plugin-registry#2042
eclipse-che/che-plugin-registry#2045
eclipse-che/che-server#812
che-incubator/configbump#206

- Also, if you are willing to explore adding s390x to your CI through IBM's GitHub Actions self-hosted runners, please let us know.

cc: @brsriniv@in.ibm.com

run/MULTIARCH_BUILD.md
```diff
@@ -2,7 +2,7 @@
 
 ## Overview
 
-The `build-multiarch.sh` script builds Docker images for multiple architectures (AMD64 and ARM64) and pushes them to a container registry.
+The `build-multiarch.sh` script builds Docker images for multiple architectures (AMD64, ARM64, and s390x) and pushes them to a container registry.
```
svor, line 5: >Myself(@swapnilsingh-ibm) and my colleague Sudharshan(@sudharshanibm) from IBM will take the responsibility for any s390x-specific failures, or investigations.

@swapnilsingh-ibm 
Could I ask you to add the information about the responsible contacts here? That way, we'll know whom to reach out to when issues arise.

.github/workflows/release.yml
akurinnoy: Thank you for clarifying the ownership - that addresses the main concern.

One more thing to consider: the release workflow now requires all three architectures to succeed. If s390x fails, no release can be made - not even for amd64 and arm64. The nightly workflow already handles this well by treating each arch as optional (`if [ -n ... ]`). Would it make sense to start with the same pattern in the release workflow and switch to mandatory once s390x builds have proven stable over a few release cycles?

Regarding self-hosted runners - that would be great. QEMU emulation for s390x is significantly slower and less reliable than native execution. If IBM can provide self-hosted s390x runners for GitHub Actions, it would resolve both the build time and flakiness concerns.

run/MULTIARCH_BUILD.md
swapnilsingh-ibm: Done, added `s390x` point of contacts to `run/MULTIARCH_BUILD.md`.

.github/workflows/release.yml
swapnilsingh-ibm: Done, `s390x` is now optional in the release manifest step, matching the nightly workflow pattern (`if [ -n "$S390X_VERSION" ]`). `amd64` and `arm64` remain mandatory. We'll switch `s390x` to mandatory once it proves stable across a few release cycles.
 
Regarding self-hosted runners — we're open to providing IBM s390x runners for GitHub Actions. Will follow up separately on that.

akurinnoy: @swapnilsingh-ibm thanks!

run/MULTIARCH_BUILD.md
svor: Thank you!

akurinnoy (approved):

PR #1643, opened 2026-07-31
feat(ai-registry): filter tools by cluster architecture

olexii4:
### What does this PR do?

Filters AI tools in `/dashboard/api/ai-registry` by the current cluster architecture so that
tools without a binary for the host platform are never surfaced to users.

Adds an optional `arch?: Architecture[]` field to `AiToolDefinition`. When set, a tool is
returned only if the cluster architecture (detected via `uname -m`, normalised to `x86_64` /
`arm64` / `s390x` / `ppc64le`) appears in the list. Tools with no `arch` field are returned on
every architecture.

**Current state of `registry.json` in [che-incubator/che-ai-tool-images@[CRW-11779](https://redhat.atlassian.net/browse/CRW-11779)](https://github.com/che-incubator/che-ai-tool-images):**

| Tool | arch constraint | Reason |
|---|---|---|
| Claude Code | `["x86_64", "arm64"]` | Anthropic publishes only `linux-x64` and `linux-arm64` binaries; no s390x/ppc64le |
| Gemini CLI |  `["x86_64", "arm64", "s390x", "ppc64le"]` | yoga WASM rebuilt with big-endian support for s390x/ppc64le |
| OpenCode | `["x86_64", "arm64"]` | Anomaly Innovations publishes only `linux-x64` and `linux-arm64` binaries; no s390x/ppc64le |

Знімок екрана 2026-08-05 о 17 05 14

fixes https://redhat.atlassian.net/browse/CRW-11779

### Is it tested? How?

1. Deploy Che with the PR images on an x86_64 cluster.
2. Run `curl https://<che-host>/dashboard/api/ai-registry | jq '[.tools[].name]'`; verify Claude Code, Gemini CLI, and OpenCode all appear.
3. Repeat on an arm64 cluster; expect the same four tools.
4. On a cluster where `uname -m` reports `s390x` : run the same `curl`; verify Claude Code is absent and the remaining three tools are present.
5. Simulate a `uname -m` failure (e.g. patch `getCurrentArchitecture` to throw); verify the endpoint still returns 200 with all tools and a `WARN` entry appears in the dashboard pod logs.

#### Release Notes

The AI tool registry endpoint now filters tools by cluster architecture; tools without a supported
binary for the host platform (for example, Claude Code on IBM Z) are automatically excluded.

#### Docs PR

https://github.com/eclipse-che/che-docs/pull/3185

#### AI Tool PR

https://github.com/che-incubator/che-ai-tool-images/pull/6


tolusha:

I check for new comments every **10m0s**, so there may be a short delay before I respond.

**Available commands**:
- `/che-ai-assistant ok-pr-readiness` — Ensure PR has validation steps
- `/che-ai-assistant check-pr-test-failures` — Analyze failing CI checks, identify root causes, and suggest fixes
- `/che-ai-assistant update-che-e2e-tests` — Update Eclipse Che e2e tests
- `/che-ai-assistant claude` — Run a free-form instruction on this PR

olexii4 (commented):
The backend architecture-filtering is clean and well-tested. The blocklist approach for event filtering is a sound default. A few things worth discussing before merge.

**Must address before merge:**
- Double-toggle in `handleVersionSelect` relies on React batching two `setState` calls in a specific order (see inline). If the parent ever adds async side effects or uses a non-functional `setState`, both calls silently cancel each other.
- `startsWith(id)` prefix check in the event selector can produce false positives when one workspace ID is a prefix of another (see inline).

**Worth addressing:**
- `componentDidUpdate` reference equality causes unnecessary re-renders - `groupByProviderId()` creates new arrays on every parent render.
- The type guard in `aiRegistryApi.ts` asserts `AiToolDefinition` shape without verifying any of its fields.
- `groupByProviderId` is duplicated across three components with inconsistent sort order.

**Suggestions (fine as follow-up):**
- No component-level test verifies that `WorkspaceEvents` correctly filters events from another workspace in the rendered output. Selector-level tests are there, but a regression where the component stops passing `workspace.id` would go undetected.
- CSS raw pixel values and inline style objects in JSX should move to CSS modules.
- The committed PR doc file describes cross-repo work not in this diff.

packages/dashboard-frontend/src/components/AiSelector/Gallery/Entry/index.tsx
```diff
@@ -72,7 +105,68 @@ export class AiProviderEntry extends React.PureComponent<Props> {
...
+    event.preventDefault();
+
+    const { isSelected, onToggle } = this.props;
+    const { activeTool } = this.state;
+
+    this.setState({ activeTool: tool, isKebabOpen: false });
+
+    // If this provider was already selected and the version changed, re-select
+    if (isSelected && activeTool.tag !== tool.tag) {
+      onToggle(tool.providerId);
```
olexii4: When the version changes while the card is already selected, `onToggle` is called twice in immediate succession. This relies on React 18 batching both `setState` calls to first deselect then re-select.

If the parent's toggle handler ever adds async side effects (an API call, a Redux thunk), two rapid calls fire two operations. A future maintainer could also "simplify" this to a single call and silently break the re-selection.

Consider a dedicated callback like `onVersionChange(providerId, newTag)` so the parent can update which tool is active directly, without abusing toggle semantics.

packages/dashboard-frontend/src/store/Events/selectors.ts
```diff
@@ -13,26 +13,56 @@
...
+          return true;
+        }
+
+        // Blocklist approach: hide events that belong to a DIFFERENT workspace.
+        // Events for the current workspace (id-prefix match or exact name match) and
+        // events not associated with any devworkspace are all kept.
+        const objName = event.involvedObject?.name ?? '';
+
+        const belongsToOtherWorkspace =
+          allWorkspaceIds.some(id => id !== currentWorkspaceId && objName.startsWith(id)) ||
```
olexii4: `objName.startsWith(id)` can match events that belong to a *different* workspace when one workspace ID is a prefix of another. For example, if IDs `workspace1abc` and `workspace1abc2def` both exist, events for `workspace1abc2def-pod-xxx` match `workspace1abc` via `startsWith`, and get incorrectly hidden when viewing `workspace1abc2def`.

Would it be possible to add a delimiter-aware check?

```suggestion
          allWorkspaceIds.some(id => id !== currentWorkspaceId && (objName === id || objName.startsWith(id + '-'))) ||
```

packages/dashboard-frontend/src/components/AiSelector/Gallery/Entry/index.tsx
```diff
@@ -27,17 +40,37 @@ export type Props = {
...
+  constructor(props: Props) {
+    super(props);
+    this.state = {
+      activeTool: props.toolGroup[0],
+      isKebabOpen: false,
+    };
+  }
+
+  public componentDidUpdate(prevProps: Props): void {
+    if (prevProps.toolGroup !== this.props.toolGroup) {
```
olexii4: `prevProps.toolGroup !== this.props.toolGroup` uses referential equality. `groupByProviderId()` in the parent (`AiProviderGallery`) creates a new array instance on every render, so every parent re-render causes `setState` in every `AiProviderEntry` - even when the tool list is unchanged.

Since `AiProviderGallery` is a `PureComponent`, this only fires when the gallery's own props change, which limits the damage. But consider adding a content comparison to be safe:
```ts
if (prevProps.toolGroup.map(t => t.tag).join(',') !== this.props.toolGroup.map(t => t.tag).join(',')) {
```
Or memoize the groups in the parent so the reference is stable when contents have not changed.

packages/dashboard-backend/src/devworkspaceClient/services/aiRegistryApi.ts
```diff
@@ -75,9 +75,16 @@ export class AiRegistryApiService implements IAiRegistryApi {
             continue;
           }
           const registry = parsed as Record<string, unknown>;
+          const rawTools = Array.isArray(registry.tools) ? (registry.tools as unknown[]) : [];
+          const allTools = rawTools.filter(
+            (t): t is api.AiToolDefinition => typeof t === 'object' && t !== null,
```
olexii4: The type predicate `t is api.AiToolDefinition` asserts the full shape, but the runtime check only verifies `typeof t === 'object' && t !== null`. Any object - `{ foo: 42 }` - passes this guard and is typed as a valid `AiToolDefinition`. Downstream code accesses `tool.arch`, `tool.providerId`, and others without null checks.

The data source is a cluster admin ConfigMap, so malformed entries are plausible. Does it make sense to add a minimal shape check, e.g.:
```ts
(t): t is api.AiToolDefinition => typeof t === 'object' && t !== null && typeof (t as Record<string, unknown>).providerId === 'string'
```

packages/dashboard-frontend/src/components/AiSelector/Gallery/Entry/index.tsx
```diff
@@ -27,17 +40,37 @@ export type Props = {
...
+type State = {
+  activeTool: api.AiToolDefinition;
+  isKebabOpen: boolean;
+};
+
+export class AiProviderEntry extends React.PureComponent<Props, State> {
+  constructor(props: Props) {
+    super(props);
+    this.state = {
+      activeTool: props.toolGroup[0],
```
olexii4, line 53: `props.toolGroup[0]` is `undefined` if an empty array is passed. The callers in `AiProviderGallery.groupByProviderId()` guarantee non-empty groups today, but `Props` does not encode that constraint. A future refactor that passes an empty array would get a runtime crash at render.

Consider using a tuple type to express the invariant:
```ts
toolGroup: [api.AiToolDefinition, ...api.AiToolDefinition[]];
```
That makes the non-empty guarantee part of the API contract and is caught at the call site.

packages/dashboard-frontend/src/components/AiSelector/Gallery/Entry/index.tsx
```diff
@@ -72,7 +105,68 @@ export class AiProviderEntry extends React.PureComponent<Props> {
...
+    this.setState({ activeTool: tool, isKebabOpen: false });
+
+    // If this provider was already selected and the version changed, re-select
+    if (isSelected && activeTool.tag !== tool.tag) {
+      onToggle(tool.providerId);
+      onToggle(tool.providerId);
+    }
+  };
+
+  private buildVersionDropdown(): React.ReactElement {
```
olexii4: The Gallery card always renders the version kebab menu, even when `toolGroup` has only one entry. In contrast, `SelectorModal.buildVersionDropdown` returns `null` for single-version providers (`toolGroup.length <= 1`).

Users who see a kebab in the gallery but nothing in the modal may be confused. Does it make sense to align the behavior, or would a short comment here explaining why the Gallery always shows it (e.g. reserved for future actions like "remove tool") help readers?

packages/dashboard-frontend/src/store/Events/selectors.ts
```diff
@@ -13,26 +13,56 @@
...
-      if (isNaN(resourceVersion) || isNaN(fromResourceVersion)) {
-        return false;
-      }
-      return fromResourceVersion <= resourceVersion;
-    });
-  };
-});
+export const selectEventsFromResourceVersion = createSelector(
+  selectAllEvents,
+  selectAllWorkspaces,
```
olexii4: Adding `selectAllWorkspaces` as a selector input means this selector recomputes whenever **any** workspace attribute changes - status updates, metadata patches, etc. `selectAllWorkspaces` calls `constructWorkspace()` for each workspace, creating new objects each time, so the memoization boundary resets on every workspace state change.

Would it be possible to depend on a lighter derived selector - one that extracts just `{ id, name }` pairs from the raw DevWorkspace state, memoized by workspace count + IDs - so the selector only recomputes when the actual workspace list changes?

packages/dashboard-frontend/src/pages/WorkspaceDetails/OverviewTab/AiTool/SelectorModal.tsx
```diff
@@ -16,41 +16,161 @@ import {
...
+    };
+  }
+
+  public componentDidUpdate(prevProps: Props): void {
+    if (prevProps.selectedVersions !== this.props.selectedVersions) {
+      this.setState({ selectedVersions: { ...this.props.selectedVersions } });
+    }
+  }
+
+  private groupByProviderId(): api.AiToolDefinition[][] {
```
olexii4: Three components now each implement their own provider-grouping logic with subtle differences: `Gallery/index.tsx` sorts before grouping (`localeCompare`), `SelectorModal` groups without sorting, and `InfoModal` uses `Array.prototype.findIndex` for dedup. A user could see tools in a different order in the gallery vs the selector modal.

Consider extracting a shared `groupToolsByProvider(tools: AiToolDefinition[]): AiToolDefinition[][]` utility in `@/services/helpers/aiTools.ts` with deterministic ordering. That would also be the right place to enforce the non-empty-group invariant and make it easier to test all three behaviours at once.

packages/dashboard-frontend/src/services/helpers/aiTools.ts
```diff
@@ -102,6 +102,33 @@ export function getInjectedAiToolIds(
...
+      container?: { image?: string };
+    }>) ?? [];
+  const versions: Record<string, string> = {};
+  for (const comp of components) {
+    const image: string = comp.container?.image ?? '';
+    // Match by stripping tag to find the provider, but keep the actual tag
+    const tool = allTools.find(t => stripImageTag(t.injectorImage) === stripImageTag(image));
+    if (tool && !versions[tool.providerId]) {
+      // Extract tag from the actual injected image
+      const tagMatch = image.match(/:([^:@]+)$/);
```
olexii4: For digest-pinned images like `quay.io/example/claude-code@sha256:abc123`, the regex `/:([^:@]+)$/` extracts `abc123` (the hash after `sha256:`) as the "version". When an admin upgrades the registry from a digest to a tagged image (or vice versa), `versionChanged()` will always see a mismatch and trigger a remove-and-re-add cycle.

This is not harmful since the operation is idempotent, but it could unexpectedly enable the Save button. Would it be worth adding a comment that digest versions are treated opaquely here, or normalizing digest references to a sentinel like `"@digest"` before comparison?

packages/dashboard-frontend/src/components/AiSelector/Gallery/Entry/index.module.css
```diff
@@ -10,6 +10,10 @@
  *   Red Hat, Inc. - initial API and implementation
  */
 
+.cardTitle {
+  padding-top: 5px;
```
olexii4: `padding-top: 5px` uses a raw pixel value. Per project conventions, PatternFly spacing tokens are preferred for consistency.

```suggestion
  padding-top: var(--pf-t--global--spacer--xs);
```

packages/dashboard-frontend/src/components/TagLabel/index.module.css
```diff
@@ -11,6 +11,9 @@
  */
 
 .label {
+  position: relative;
+  top: 2px;
```
olexii4: `top: 2px` uses a raw pixel value. Consider replacing with a PatternFly token or `0.125rem` to stay consistent with the rest of the stylesheet.

packages/dashboard-frontend/src/components/AiSelector/Gallery/Entry/index.tsx
```diff
@@ -113,25 +209,38 @@ export class AiProviderEntry extends React.PureComponent<Props> {
...
-            )}
-            {provider.name}
-            {provider.envVarName && hasExistingKey && (
-              <Badge isRead style={{ whiteSpace: 'nowrap' }}>
-                <CheckCircleIcon /> Key configured
-              </Badge>
-            )}
+            <div>{activeTool.name}</div>
+            <span
+              style={{
```
olexii4: This `style={{...}}` object literal is recreated on every render, creating a new object reference each time. Moving these properties to `index.module.css` would be consistent with the rest of the component and avoids unnecessary allocations.

packages/dashboard-frontend/openspec/docs/pr-CRW-11779-ai-tools-s390x-arch-filter.md
```diff
@@ -0,0 +1,66 @@
+# feat(ai): arch-aware tool filtering, version selector, and s390x/ppc64le OpenCode support
```
olexii4: This doc file mentions the `opencode:insiders` Dockerfile in `che-ai-tool-images` and the s390x/ppc64le `registry.json` changes - work that lives in a companion repo and is not reviewable here. Reviewers reading this may be confused about the scope of this PR.

Would it make sense to trim the doc to cover only what is in this diff, or add a note pointing to the companion PR for the cross-repo parts?

packages/dashboard-frontend/src/components/AiSelector/Gallery/Entry/index.tsx
```diff
@@ -72,7 +105,72 @@ export class AiProviderEntry extends React.PureComponent<Props> {
...
+    event.preventDefault();
+
+    const { isSelected, onToggle } = this.props;
+    const { activeTool } = this.state;
+
+    this.setState({ activeTool: tool, isKebabOpen: false });
+
+    // If this provider was already selected and the version changed, re-select
+    if (isSelected && activeTool.tag !== tool.tag) {
+      onToggle(tool.providerId);
```
olexii4: When the selected card's version changes, `onToggle` is called twice in immediate succession to deselect then re-select. This relies on the parent treating toggle as a symmetric flip, which is an implementation detail. If any parent adds analytics, async effects, or side effects to the toggle handler, those fire twice - and the intent (version change) is not communicated to the parent at all.

Consider a dedicated `onVersionChange(providerId: string, tag: string)` callback instead. That would communicate intent explicitly without coupling to toggle semantics.

packages/dashboard-frontend/src/components/AiSelector/Gallery/Entry/index.tsx
olexii4: `prevProps.toolGroup !== this.props.toolGroup` is a reference check. The parent calls `groupToolsByProvider()` in `render()`, which returns a new array instance on every render. Any parent re-render - even one unrelated to the tool list - creates a new array and triggers `setState` in every `AiProviderEntry`, even when the array contents are identical.

Consider comparing a stable derived value, e.g.:
```ts
if (prevProps.toolGroup.map(t => t.tag).join(',') !== this.props.toolGroup.map(t => t.tag).join(',')) {
```
Or memoize the groups in the parent so the reference is stable when the underlying tools have not changed.

packages/dashboard-frontend/src/pages/WorkspaceDetails/OverviewTab/AiTool/InfoModal.tsx
```diff
@@ -49,19 +49,23 @@ export class AiToolInfoModal extends React.PureComponent<Props> {
...
-                          {def.name}
-                        </a>
-                      </b>
-                      {provider?.description && <> — {provider.description}</>}
-                    </Content>
-                  );
-                })}
+                {aiTools
+                  .filter(
+                    (def, i, arr) => arr.findIndex(d => d.providerId === def.providerId) === i,
```
olexii4: The PR extracted `groupToolsByProvider()` to `aiTools.ts` and updated Gallery and SelectorModal to use it. InfoModal still uses an inline `findIndex` for dedup. If the dedup logic in `groupToolsByProvider` is later updated, InfoModal would silently diverge.

Consider using the shared utility here too:
```ts
import { groupToolsByProvider } from '@/services/helpers/aiTools';
// ...
{groupToolsByProvider(aiTools).map(toolGroup => {
  const def = toolGroup[0];
```

packages/dashboard-frontend/src/pages/WorkspaceDetails/OverviewTab/AiTool/SelectorModal.tsx
```diff
@@ -62,24 +177,42 @@ export class AiToolSelectorModal extends React.PureComponent<Props> {
...
-                {aiTools.map(def => {
-                  const provider = aiProviders.find(p => p.id === def.providerId);
+                {groups.map(toolGroup => {
+                  const providerId = toolGroup[0].providerId;
+                  const activeTag = selectedVersions[providerId] ?? toolGroup[0].tag;
+                  const provider = aiProviders.find(p => p.id === providerId);
+                  const versionDropdown = this.buildVersionDropdown(toolGroup);
+
+                  const checkboxLabel = (
+                    <span style={{ display: 'inline-flex', alignItems: 'center', gap: '4px' }}>
```
olexii4: `style={{ display: 'inline-flex', alignItems: 'center', gap: '4px' }}` is recreated on every render. A similar span was moved to a CSS class (`badgeGroup`) in `Entry/index.module.css`. Would it be possible to do the same here with a CSS module class in SelectorModal? Also, line 200 has `minWidth: '45px'` and `height: '13px'` as raw pixel values inside an inline style - those could use PatternFly spacing tokens or rem values.

packages/dashboard-frontend/src/services/helpers/aiTools.ts
```diff
@@ -196,8 +228,11 @@ export function addAiToolToWorkspace(
   workspaceOrDevWorkspace: Workspace | devfileApi.DevWorkspace,
   toolId: string,
   allTools: api.AiToolDefinition[],
+  tag?: string,
```
olexii4, line 231: The new optional `tag` parameter selects which version of a tool gets injected. The tag-based lookup logic is non-trivial:
```ts
const tool =
  (tag ? allTools.find(t => t.providerId === toolId && t.tag === tag) : undefined) ??
  allTools.find(t => t.providerId === toolId);
```
No test in `aiTools.spec.ts` covers this path. Missing cases:
- Passing a valid `tag` that matches a specific version (should prefer that version)
- Passing an unknown `tag` (should fall back to first match by `providerId`)
- Passing `undefined` (should preserve existing behavior)

The version selection feature relies on this logic to inject the correct binary into the workspace.

packages/dashboard-frontend/src/pages/WorkspaceDetails/OverviewTab/AiTool/index.tsx
```diff
@@ -71,37 +80,50 @@ class AiToolFormGroup extends React.PureComponent<Props, State> {
...
+
+    const noSelectionChange = selected.join(',') === currentToolIds.join(',');
+    const noVersionChange = selected.every(id => !versionChanged(id));
+
+    if (noSelectionChange && noVersionChange) {
       this.setState({ isSelectorOpen: false });
       return;
     }
 
     let updatedDw = cloneDeep(workspace.ref);
```
olexii4: `let updatedDw = cloneDeep(workspace.ref)` creates a deep copy before entering the mutation loops. Both `removeAiToolFromWorkspace` and `addAiToolToWorkspace` internally do `JSON.parse(JSON.stringify(raw))`, so the first call re-clones the already-cloned value. The outer `cloneDeep` is redundant - passing `workspace.ref` directly would work the same way and avoids an unnecessary full clone.

packages/dashboard-frontend/src/services/helpers/aiTools.ts
```diff
@@ -102,6 +102,38 @@ export function getInjectedAiToolIds(
   return ids;
 }
 
+/**
+ * Returns a map of providerId → injected tag for all injected AI tools.
+ * Used to detect which version of a tool is currently in the workspace.
+ */
+export function getInjectedAiToolVersions(
```
olexii4, line 130: `getInjectedAiToolIds` and `getInjectedAiToolVersions` iterate the same component list with the same `stripImageTag` matching expression. They are called as a pair in `AiToolFormGroup` across five lifecycle methods - 10 calls total. If the matching logic is updated in one function but not the other, they could silently disagree about which tools are injected.

Consider unifying them into a single function:
```typescript
function getInjectedAiToolInfo(
  workspace: Workspace,
  allTools: api.AiToolDefinition[],
): { ids: string[]; versions: Record<string, string> } { ... }
```
Callers destructure as needed: `const { ids, versions } = getInjectedAiToolInfo(...)`.

packages/dashboard-frontend/src/pages/WorkspaceDetails/OverviewTab/AiTool/__tests__/SelectorModal.spec.tsx
```diff
@@ -72,7 +72,9 @@ function getComponent(
       aiTools={tools}
       aiProviders={providers}
       selected={selected}
+      selectedVersions={{}}
```
olexii4, line 84: `selectedVersions` and `originVersions` props were added to `getComponent()`, but no new test cases cover the version-selection feature. `AiToolSelectorModal` is the primary path for users to change tool versions on existing workspaces.

Missing cases to consider:
- Save button enabled when only the version changes (not the selection)
- Version dropdown rendering for multi-version tool groups
- Auto-selection when a version is picked for an unselected tool
- `onConfirm` called with the correct `selectedVersions` map
- `hasVersionChange` computation correctness

olexii4 (commented):
## General findings

### AiSelector factory flow ignores the user's version choice

`AiSelector.handleProviderToggle` tracks only `selectedProviderIds: string[]`. The version dropdown in `AiProviderEntry` updates local component state (`activeTool`) but that choice is never surfaced to `AiSelector` - it only calls `onSelect(providerIds)`. Callers then use `addAiToolToWorkspace(workspace, toolId, aiTools)` without a `tag` argument, always picking the first tool matching `providerId` regardless of what the user selected.

The version dropdown in the factory/create-workspace gallery is effectively cosmetic. `SelectorModal` (the workspace-details path) correctly propagates versions via `onConfirm(selectedVersions)`, but the `AiSelector` path does not.

Does it make sense to add a `selectedVersions` map to `AiSelector` state and thread it through `onSelect(providerIds, selectedVersions)` - similar to how `AiToolFormGroup` handles it?

---

### WorkspaceEvents component test does not verify workspace-scoped filtering

The `selectors.spec.ts` tests cover the filtering logic well. However, there are no component-level tests in `WorkspaceEvents/__tests__/index.spec.tsx` that verify events from another workspace are hidden in the rendered output. A regression where the component stops passing `workspace.id` to `eventsFromResourceVersionFn` would not be caught at the selector level.

Would it be possible to add one test that renders `WorkspaceEvents` with two workspaces in the store and checks that events from the other workspace do not appear?

packages/dashboard-frontend/src/components/AiSelector/Gallery/Entry/index.tsx
olexii4: When the selected card's version changes, `onToggle` is called twice in immediate succession to deselect then re-select. This relies on the parent treating toggle as a symmetric flip, which is an implementation detail. If any parent adds analytics, async effects, or side effects to the toggle handler, those fire twice - and the intent (version change) is not communicated to the parent at all.

Consider a dedicated `onVersionChange(providerId: string, tag: string)` callback instead. That would communicate intent explicitly without coupling to toggle semantics.

packages/dashboard-frontend/src/components/AiSelector/Gallery/Entry/index.tsx
olexii4: `prevProps.toolGroup !== this.props.toolGroup` is a reference check. The parent calls `groupToolsByProvider()` in `render()`, which returns a new array instance on every render. Any parent re-render - even one unrelated to the tool list - creates a new array and triggers `setState` in every `AiProviderEntry`, even when the array contents are identical.

Consider comparing a stable derived value, e.g.:
```ts
if (prevProps.toolGroup.map(t => t.tag).join(',') !== this.props.toolGroup.map(t => t.tag).join(',')) {
```
Or memoize the groups in the parent so the reference is stable when the underlying tools have not changed.

packages/dashboard-frontend/src/pages/WorkspaceDetails/OverviewTab/AiTool/InfoModal.tsx
olexii4: The PR extracted `groupToolsByProvider()` to `aiTools.ts` and updated Gallery and SelectorModal to use it. InfoModal still uses an inline `findIndex` for dedup. If the dedup logic in `groupToolsByProvider` is later updated, InfoModal would silently diverge.

Consider using the shared utility here too:
```ts
import { groupToolsByProvider } from '@/services/helpers/aiTools';
// ...
{groupToolsByProvider(aiTools).map(toolGroup => {
  const def = toolGroup[0];
```

packages/dashboard-frontend/src/pages/WorkspaceDetails/OverviewTab/AiTool/SelectorModal.tsx
olexii4: `style={{ display: 'inline-flex', alignItems: 'center', gap: '4px' }}` is recreated on every render. A similar span was moved to a CSS class (`badgeGroup`) in `Entry/index.module.css`. Would it be possible to do the same here with a CSS module class in SelectorModal? Also, line 200 has `minWidth: '45px'` and `height: '13px'` as raw pixel values inside an inline style - those could use PatternFly spacing tokens or rem values.

packages/dashboard-frontend/src/services/helpers/aiTools.ts
olexii4, line 231: The new optional `tag` parameter selects which version of a tool gets injected. The tag-based lookup logic is non-trivial:
```ts
const tool =
  (tag ? allTools.find(t => t.providerId === toolId && t.tag === tag) : undefined) ??
  allTools.find(t => t.providerId === toolId);
```
No test in `aiTools.spec.ts` covers this path. Missing cases:
- Passing a valid `tag` that matches a specific version (should prefer that version)
- Passing an unknown `tag` (should fall back to first match by `providerId`)
- Passing `undefined` (should preserve existing behavior)

The version selection feature relies on this logic to inject the correct binary into the workspace.

packages/dashboard-frontend/src/pages/WorkspaceDetails/OverviewTab/AiTool/index.tsx
olexii4: `let updatedDw = cloneDeep(workspace.ref)` creates a deep copy before entering the mutation loops. Both `removeAiToolFromWorkspace` and `addAiToolToWorkspace` internally do `JSON.parse(JSON.stringify(raw))`, so the first call re-clones the already-cloned value. The outer `cloneDeep` is redundant - passing `workspace.ref` directly would work the same way and avoids an unnecessary full clone.

packages/dashboard-frontend/src/services/helpers/aiTools.ts
olexii4, line 130: `getInjectedAiToolIds` and `getInjectedAiToolVersions` iterate the same component list with the same `stripImageTag` matching expression. They are called as a pair in `AiToolFormGroup` across five lifecycle methods - 10 calls total. If the matching logic is updated in one function but not the other, they could silently disagree about which tools are injected.

Consider unifying them into a single function:
```typescript
function getInjectedAiToolInfo(
  workspace: Workspace,
  allTools: api.AiToolDefinition[],
): { ids: string[]; versions: Record<string, string> } { ... }
```
Callers destructure as needed: `const { ids, versions } = getInjectedAiToolInfo(...)`.

packages/dashboard-frontend/src/pages/WorkspaceDetails/OverviewTab/AiTool/__tests__/SelectorModal.spec.tsx
olexii4, line 84: `selectedVersions` and `originVersions` props were added to `getComponent()`, but no new test cases cover the version-selection feature. `AiToolSelectorModal` is the primary path for users to change tool versions on existing workspaces.

Missing cases to consider:
- Save button enabled when only the version changes (not the selection)
- Version dropdown rendering for multi-version tool groups
- Auto-selection when a version is picked for an unselected tool
- `onConfirm` called with the correct `selectedVersions` map
- `hasVersionChange` computation correctness

olexii4 (commented):
## General findings

### AiSelector factory flow ignores the user's version choice

`AiSelector.handleProviderToggle` tracks only `selectedProviderIds: string[]`. The version dropdown in `AiProviderEntry` updates local component state (`activeTool`) but that choice is never surfaced to `AiSelector` - it only calls `onSelect(providerIds)`. Callers then use `addAiToolToWorkspace(workspace, toolId, aiTools)` without a `tag` argument, always picking the first tool matching `providerId` regardless of what the user selected.

The version dropdown in the factory/create-workspace gallery is effectively cosmetic. `SelectorModal` (the workspace-details path) correctly propagates versions via `onConfirm(selectedVersions)`, but the `AiSelector` path does not.

Does it make sense to add a `selectedVersions` map to `AiSelector` state and thread it through `onSelect(providerIds, selectedVersions)` - similar to how `AiToolFormGroup` handles it?

---

### WorkspaceEvents component test does not verify workspace-scoped filtering

The `selectors.spec.ts` tests cover the filtering logic well. However, there are no component-level tests in `WorkspaceEvents/__tests__/index.spec.tsx` that verify events from another workspace are hidden in the rendered output. A regression where the component stops passing `workspace.id` to `eventsFromResourceVersionFn` would not be caught at the selector level.

Would it be possible to add one test that renders `WorkspaceEvents` with two workspaces in the store and checks that events from the other workspace do not appear?

packages/dashboard-frontend/src/store/Events/selectors.ts
```diff
@@ -13,26 +13,68 @@
...
-    });
-  };
-});
+// Lightweight selector that extracts only {id, name} from raw DevWorkspace state.
+// This avoids constructing full Workspace adapter objects on every workspace status
+// update, so selectEventsFromResourceVersion only recomputes when workspace
+// identifiers actually change (not on every status patch).
+const selectWorkspaceIdNamePairs = createSelector(selectAllDevWorkspaces, devWorkspaces =>
+  devWorkspaces.map(dw => ({
+    id: dw.status?.devworkspaceId ?? 'workspace' + dw.metadata.uid.split('-').splice(0, 3).join(''),
```
akurinnoy: `splice` mutates the array it's called on. Harmless here since `split('-')` returns a throwaway array, but it still makes a pure selector look suspicious - a reader has to stop and check whether shared state got touched. `slice(0, 3)` does the same thing and reads correctly as non-mutating.

```suggestion
    id: dw.status?.devworkspaceId ?? 'workspace' + dw.metadata.uid.split('-').slice(0, 3).join(''),
```

packages/dashboard-frontend/src/pages/WorkspaceDetails/OverviewTab/AiTool/index.tsx
```diff
@@ -38,24 +40,28 @@ export type State = {
...
-    const newToolIds = getInjectedAiToolIds(workspace, aiTools);
-    const prevToolIds = getInjectedAiToolIds(prevProps.workspace, prevProps.aiTools);
+    const { ids: newToolIds } = getInjectedAiToolInfo(workspace, aiTools);
+    const { ids: prevToolIds } = getInjectedAiToolInfo(prevProps.workspace, prevProps.aiTools);
     if (newToolIds.join(',') !== prevToolIds.join(',')) {
-      this.setState({ selected: newToolIds });
+      const { ids, versions } = getInjectedAiToolInfo(workspace, aiTools);
+      this.setState({ selected: ids, selectedVersions: versions });
     }
   }
```
akurinnoy, line 65: `getInjectedAiToolInfo(workspace, aiTools)` is called twice with the same arguments here - once to compare IDs, again to get the result when they differ. The first call already has what you need.

```suggestion
  public componentDidUpdate(prevProps: Props): void {
    const { aiTools, workspace } = this.props;
    const { ids: newToolIds, versions: newVersions } = getInjectedAiToolInfo(workspace, aiTools);
    const { ids: prevToolIds } = getInjectedAiToolInfo(prevProps.workspace, prevProps.aiTools);
    if (newToolIds.join(',') !== prevToolIds.join(',')) {
      this.setState({ selected: newToolIds, selectedVersions: newVersions });
    }
  }
```

packages/dashboard-frontend/src/components/AiSelector/index.tsx
```diff
@@ -218,6 +218,11 @@ class AiSelector extends React.PureComponent<Props, State> {
                           selectedProviderIds={selectedProviderIds}
                           providerKeyExists={providerKeyExists}
                           onToggle={providerId => this.handleProviderToggle(providerId)}
+                          onVersionChange={() => {
+                            // Version change in the gallery — no action needed here since
+                            // this selector is for provider selection only. Version state
+                            // is managed by SelectorModal for existing workspaces.
```
akurinnoy: `onVersionChange` is a no-op here. Pick a version from the gallery card before creating a workspace and the tag updates visually, but the choice is silently dropped - only `SelectorModal` (workspace details) actually persists it.

Does it make sense to hide the dropdown here, or add a note that version can only be changed after the workspace exists? Right now it looks functional but does nothing.

packages/dashboard-frontend/src/pages/WorkspaceDetails/OverviewTab/AiTool/index.tsx
```diff
@@ -69,39 +75,53 @@ class AiToolFormGroup extends React.PureComponent<Props, State> {
...
+
+    if (noSelectionChange && noVersionChange) {
       this.setState({ isSelectorOpen: false });
       return;
     }
 
-    let updatedDw = cloneDeep(workspace.ref);
+    // removeAiToolFromWorkspace / addAiToolToWorkspace each clone internally,
+    // so passing workspace.ref directly avoids a redundant up-front deep copy.
+    let updatedDw = workspace.ref;
```
akurinnoy, line 107: Good catch dropping the redundant `cloneDeep` - `removeAiToolFromWorkspace`/`addAiToolToWorkspace` already clone internally, so this is correct today. But it's now an implicit contract: if either function ever starts mutating in place, `workspace.ref` from the Redux store gets silently mutated.

Worth naming that contract explicitly:

```suggestion
    // SAFETY: relies on removeAiToolFromWorkspace / addAiToolToWorkspace cloning workspace.ref
    // internally via JSON.parse(JSON.stringify(...)). If those functions change to mutate
    // in place, this line must be reverted to cloneDeep(workspace.ref).
    let updatedDw = workspace.ref;
```

packages/dashboard-frontend/openspec/docs/pr-CRW-11779-ai-tools-s390x-arch-filter.md
```diff
@@ -0,0 +1,68 @@
...
```
akurinnoy: Not a blocker, but this file describes cross-repo work (Dockerfiles and `registry.json` changes in `che-ai-tool-images`) that isn't part of this diff. Future readers finding it in a committed path won't know if it's current design or a superseded proposal.

I think this belongs in the PR description, not committed to the repo. Or at least add a header noting it's historical.

packages/dashboard-frontend/src/store/Events/selectors.ts
olexii4: I have fixed it.

packages/dashboard-frontend/src/pages/WorkspaceDetails/OverviewTab/AiTool/index.tsx
olexii4: Fixed. The redundant call to getInjectedAiToolInfo in componentDidUpdate was
  removed — ids and versions are now destructured from the first call and reused
  directly in setState.

packages/dashboard-frontend/openspec/docs/pr-CRW-11779-ai-tools-s390x-arch-filter.md
olexii4: I have fixed it.

packages/dashboard-frontend/src/components/AiSelector/index.tsx
olexii4: I have fixed it.

packages/dashboard-frontend/src/pages/WorkspaceDetails/OverviewTab/AiTool/index.tsx
olexii4: Done. Updated the comment to your exact suggestion — **SAFETY** prefix, the `JSON.parse(JSON.stringify(...)) ` mechanism named explicitly, and the revert instruction in place.

svor (approved):
LGTM

Image

Image

PR #1649, opened 2026-08-10
fix: include devfilePath in workspace reuse check

olexii4:
### What does this PR do?

Fixes workspace reuse incorrectly ignoring `devfilePath` when checking for an existing workspace from the same repository.

#### Root Cause

`getSameRepoWorkspaces()` in `CheckExistingWorkspaces` matched workspaces using only the source URL and revision. When a factory link included `?devfilePath=apps/appA.yaml`, the parameter was forwarded to the factory resolver as `override.devfileFilename` but was never stored in `FactoryParams` for deduplication purposes. As a result, opening a second factory link pointing to a different devfile in the same repo — e.g. `?devfilePath=apps/appB.yaml` — matched the workspace from `appA.yaml` and reopened it instead of creating a new one.

#### Fix

Adds a `devfilePath` getter to `WorkspaceAdapter` that reads `override.devfileFilename` from the `factory.params` annotation stored on each DevWorkspace. The `getSameRepoWorkspaces()` filter now includes this value in its equality check alongside source URL and revision.

```diff
-      return workspace.source === factoryParams.sourceUrl && revision === factoryParams.revision;
+      return (
+        workspace.source === factoryParams.sourceUrl &&
+        revision === factoryParams.revision &&
+        workspace.devfilePath === factoryDevfilePath
+      );
```

Two workspaces created from the same repo are now considered distinct when they were started from different devfiles, and the same workspace is still correctly reused when the source URL, revision, and devfile path all match.

N/A

fixes https://redhat.atlassian.net/browse/CRW-10861

### Is it tested? How?

1. Deploy Che with the dashboard image from this PR.
2. Open a factory link for devfile A in a central repo:
   ```
   https://<che-host>#https://github.com/org/central-repo.git?devfilePath=apps/appA/devfile.yaml
   ```
   Verify workspace A is created and starts successfully.
3. Open a factory link for devfile B in the same repo:
   ```
   https://<che-host>#https://github.com/org/central-repo.git?devfilePath=apps/appB/devfile.yaml
   ```
   Verify a new workspace B is created — workspace A is **not** reopened.
4. Open the factory link for devfile A again; verify workspace A is reopened (correct reuse still works).

#### Release Notes

Fixed an issue where Dev Spaces would reopen an existing workspace instead of creating a new one when two factory links pointed to different devfiles in the same repository.

#### Docs PR

N/A

### DIFF
--- a/packages/dashboard-frontend/src/components/WorkspaceProgress/CreatingSteps/CheckExistingWorkspaces/__tests__/index.spec.tsx
+++ b/packages/dashboard-frontend/src/components/WorkspaceProgress/CreatingSteps/CheckExistingWorkspaces/__tests__/index.spec.tsx
@@ -831,6 +831,96 @@ describe('Creating steps, checking existing workspaces', () => {
         await waitFor(() => expect(mockOnNextStep).toHaveBeenCalled());
         expect(mockTabManagerReplace).not.toHaveBeenCalled();
       });
+
+      it('should reuse a workspace created without override when the factory URL specifies the resolver default source', async () => {
+        const defaultSource = 'devfile.yaml';
+        const localSearchParams = new URLSearchParams({
+          [FACTORY_URL_ATTR]: factoryUrl,
+          'override.devfileFilename': defaultSource,
+        });
+        // Workspace was created without override.devfileFilename (devfilePath = undefined)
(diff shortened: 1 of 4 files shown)


svor (approved):
lgtm 

Image

PR #1651, opened 2026-08-14
A11y container registry aria labels

bianicu:
### What does this PR do?
This PR improves aria-labels in the Container Registry modal to include programmatic context for screen     
  reader users.                                                                            

  It updates two buttons with generic labels to have more descriptive ones:                                   
   
  1. Open registry URL button — changes aria-label="open registry" → aria-label="Open container registry"     
  2. Show password toggle button — changes aria-label="show" → aria-label="Show password"
                                                                                                              
  This makes these interactive elements more accessible when users navigate with screen readers or use        
  assistive technology, as the labels now clearly describe the button's purpose even when read out of visual
  context.                                                                                      

https://redhat.atlassian.net/browse/CRW-10285


packages/dashboard-frontend/openspec/docs/pr-a11y-container-registry-aria-labels.md
```diff
@@ -0,0 +1,35 @@
+### What does this PR do?
```
svor: @bianicu We don't need this pr-a11y-container-registry-aria-labels.md‎

olexii4:
The two prop fixes are clean — `Show password` and `Open container registry` are good labels.

The close button approach needs a change. `updateCloseButtonAriaLabel` uses an unscoped `document.querySelector`:

```ts
document.querySelector('.pf-v6-c-modal-box .pf-v6-c-modal-box__close button')
```

Two problems:

1. **Wrong target if two modals are open.** The selector matches the first \`.pf-v6-c-modal-box\` in the document, not necessarily this component's modal. There's only one Container Registry modal at a time, but the pattern is fragile by design.

2. **Silently overwritten on re-render.** `setAttribute` writes to the live DOM node. When PatternFly re-renders `ModalContent` for any reason, it replaces the DOM node with a fresh one carrying `aria-label="Close"` and the mutation is lost. This is also why the `componentDidMount` call is a no-op when the modal starts closed — PatternFly returns `null` when `isOpen={false}`, so there's no button to find.

A declarative alternative that avoids both problems: omit `onClose` from `<Modal>` (which stops PF from rendering its default close button) and add `onEscapePress` so ESC still works, then render the close button explicitly as a child:

```diff
 <Modal
   variant={ModalVariant.small}
   isOpen={isOpen}
-  onClose={onCancel}
+  onEscapePress={onCancel}
   aria-label="edit-registry-info"
   elementToFocus="[data-pf-initial-focus]"
 >
+  <div className="pf-v6-c-modal-box__close">
+    <Button
+      variant="plain"
+      aria-label="Close Container Registry form"
+      onClick={onCancel}
+      icon={<TimesIcon />}
+    />
+  </div>
   <ModalHeader title={modalTitle} />
```

PF's CSS positions `.pf-v6-c-modal-box__close` absolutely at the top-right of the modal box regardless of where it appears in `children`, so the layout is identical. `onEscapePress` is called by PF's ESC handler when set, falling back to `onClose` only when it's absent — so keyboard dismissal still works. The `componentDidMount`/`componentDidUpdate`/`updateCloseButtonAriaLabel` methods can all be removed.

This also makes the label testable in unit tests — it's a plain prop on a button your component owns, not a DOM side effect.

olexii4 (approved):
Image

Good fixes. Three separate improvements bundled cleanly:
  
  - "Show password" and "Open container registry" — both more descriptive than the
  generic "show" / "open registry" labels that were there before.
  - The DOM workaround for the modal close button is the right call — PF6 doesn't expose
  a prop for this, and the selector .pf-v6-c-modal-box__close button is stable enough for
  the current PF version. Worth dropping a comment if it breaks on a PF upgrade.

  LGTM

SkorikSergey (approved):
Looks good to merge. The updated aria-label values match the issue requirements and provide clearer context for screen reader users.

svor (approved):

PR #1656, opened 2026-09-01
fix: prevent Create New switch reset when URL is entered in ImportFromGit

olexii4:
### What does this PR do?

Fixes the "Create New" switch being silently reset to OFF when a user enters a URL in the Import from Git field, causing a second workspace from the same repository URL to redirect to the existing one instead of creating a new one.

### Root Cause

Two cooperating bugs caused the switch to reset every time a valid Git URL was typed.

**1. `CreateNewIfExistSwitch` never wrote its initial state to `Navigation.pageState`.**

`CreateNewIfExistSwitch.componentDidMount` called `props.onChange(true)` (propagating the default ON state to the samples list), but it never set `Navigation.pageState[CREATE_NEW_IF_EXIST_SWITCH_ID]`. The static store stayed at `{ isChecked: undefined }`.

**2. `CreateNewIfExistingField.componentDidMount` unconditionally overwrote `Navigation.pageState` with `false`.**

When the URL field validated successfully, `RepoOptionsAccordion` mounted, which in turn mounted the hidden `CreateNewIfExistingField`. Its `componentDidMount` ran:

```typescript
Navigation.pageState[CREATE_NEW_IF_EXIST_SWITCH_ID] = {
  isChecked: this.state.createNewIfExisting, // always false — the field default
};
```

Because `Navigation.pageState` was `undefined` (bug 1), this write was never blocked. The setter broadcast `false` to all subscribers, including `CreateNewIfExistSwitch`, which flipped the visible switch to OFF and called `handleCreateNewIfExistChange(false)`. The URL never received the `?new` query param, so the factory loader used `policies.create=peruser`, and `CheckExistingWorkspaces` silently redirected to the existing workspace.

N/A

fixes https://redhat.atlassian.net/browse/CRW-12731

### Is it tested? How?

2. Navigate to **Create Workspace** (Get Started page). Confirm the **Create New** switch in the toolbar is ON.
3. Enter a public GitHub or GitLab URL (e.g. `https://github.com/che-incubator/quarkus-api-example`) and click **Create & Open**. Workspace A is created.
4. Return to the same page, confirm the **Create New** switch is still ON. Enter the same URL and click **Create & Open** — a second, distinct workspace B must be created (not a redirect to A).
5. Toggle **Create New** OFF. Enter the same URL and click **Create & Open** — the dashboard must navigate to the existing workspace A instead of creating a new one.

#### Release Notes

Fixed an issue where creating a second workspace from the same Git repository URL would reopen the existing workspace instead of creating a new one.

#### Docs PR

N/A


svor:
@SkorikSergey please validate

SkorikSergey:
@olexii4 hello, still can reproduce this issue with https://github.com/prabhuk25/PublicRepo repo. But works well with our test githup repos. The only difference I see - metadata name. In prabhuk25 repo it is `Devspaces-Public-repo`. Works as expected after changing it to `devspaces-public-repo` (https://github.com/SkorikSergey/PublicRepo/blob/main/devfile.yaml).

SkorikSergey:
@olexii4 also there is e2e tests regression - after setting url to `Git repo URL ` field and clicking on `Create & Open` button there is only part of url used as factory url(CreateWorkspaceWithExistingNameFromGitUrl test).  Cannot reproduce manually.
Screenshot

olexii4:
@SkorikSergey fixed. The regression had two cooperating causes.

**Root cause:**

`CreateNewIfExistingField.componentDidMount` called `onChange(true)` when it adopted the global switch state. That propagated through `AdvancedOptions` → `RepoOptionsAccordion` → `ImportFromGit`, which encoded `?new` into `state.location` and then overwrote the input's DOM value with `url?new`. When the user clicked "Create & Open", `buildFactoryLoaderPath` correctly moved `?new` → `policies.create=perclick`, but the factory `url` param contained only the clean git URL — the E2E test compared the input value (`url?new`) against the factory `url` param (`url`) and reported "only part of url used".

Also, `componentDidUpdate` compared `prevProps.createNewIfExisting` against `undefined || false` instead of the raw prop value, which immediately overwrote the silently-adopted local state.

**Fix:**

1. `componentDidMount` now adopts global switch state via `setState` only — no `onChange`, so the URL and input field are not touched on mount.
2. `componentDidUpdate` now compares raw prop values (`prevProps.createNewIfExisting !== this.props.createNewIfExisting`) to avoid the spurious reset.
3. `ImportFromGit.startFactory()` reads `Navigation.pageState` at click-time and injects `?new` directly before calling `buildFactoryLoaderPath`, so `policies.create=perclick` is still applied when the switch is ON.

Could you check again with https://github.com/prabhuk25/PublicRepo?

SkorikSergey (approved):
Looks good to merge

PR #1629, opened 2026-07-07
fix(backend): fix kubeconfig injection race in slow-starting workspaces

olexii4:
### What does this PR do?

Fixes two race conditions in `PostStartInjector` that caused user OAuth tokens to not be
injected into workspace terminals intermittently after PR #1606.

#### Root Cause

**Race 1 — Already-RUNNING workspace (primary):**
`watchAndInject()` starts a Kubernetes watch and waits for a `RUNNING` event. The watch
protocol does a LIST then opens a streaming WATCH from that list's `resourceVersion`. A
fast-starting workspace can reach `Running` inside the narrow window between LIST and the
stream opening — the `RUNNING` event falls before the stream's start boundary and is never
delivered. With no watch event and no polling yet, injection never happens.

**Race 2 — Polling fallback too slow (secondary):**
When the watch falls back to polling, the interval was 10 s. The UDI `entrypoint.sh`
polls for `~/.kube/config` for ~12 s before giving up. A workspace becoming `Running`
right after a poll check left a 10 s gap — long enough for the container to give up and
fall back to the pod service account identity.

Both races are observed in the CI log from 2026-07-07 as `KubedockPodmanTest` failures:
```
Kubeconfig doesn't exist yet. Waiting...  [repeated 10×]
Could not find Kubeconfig at /home/user/.kube/config
Giving up...
Error: logging into "image-registry...": invalid username/password
```

#### Fix

1. **Immediate initial check** — after registering the watch, a single `getByName()` is
   issued immediately. If the workspace is already `Running`, credentials are injected
   without waiting for watch events, eliminating the LIST→STREAM race window.

2. **Reduced poll interval** — `POLL_INTERVAL_MS` reduced from `10 000 ms` → `2 000 ms`,
   giving ~6 poll attempts within the 12 s window the UDI entrypoint.sh waits.

### Screenshot/screencast of this PR

N/A — backend-only change.

### What issues does this PR fix or reference?

fixes https://redhat.atlassian.net/browse/CRW-11193

### Is it tested? How?

1. Deploy Eclipse Che with the dashboard image from this PR.
2. Start a workspace from a factory URL (fast-path: workspace reaches `Running` before
   the watch stream opens).
3. Open a terminal and run `oc whoami` — verify it returns the OpenShift user identity,
   not `system:serviceaccount:<ns>:<sa>`.
4. Run `oc whoami -t` — verify a valid OAuth token is returned.

#### Release Notes

Fixed intermittent failure where workspace terminals used the pod service account identity
[...]

PR #1606, opened 2026-06-18
fix(backend): restore user OAuth token injection in workspace terminals (CRW-11193)

olexii4:
### What does this PR do?
This PR fixes `oc whoami` returning the workspace pod service account identity instead of the logged-in
user's identity in workspace terminals after Dev Spaces 3.28.1.

### Root Cause

PR [#1475](https://github.com/eclipse-che/che-dashboard/pull/1475) moved kubeconfig injection from
the frontend WebSocket handler to a backend `PostStartInjector` that sets up a Kubernetes Watch
when a workspace is started (PATCH or POST with `spec.started=true`).

Before #1475, the backend's own service account performed the watch (via the existing WebSocket
infrastructure), and the user's token was only needed at injection time — sourced from the
frontend's POST request to `/namespace/:ns/devworkspaceId/:id/kubeconfig`. This difference in
watch identity is what introduced the regression.

### Fix

Two changes in sequence:

**1. Reuse `watchInNamespace()` — same code path as WebSocket `SUBSCRIBE DEV_WORKSPACE`.**
The standalone `k8s.Watch` is replaced with `devworkspaceApi.watchInNamespace()`, the exact same
method the WebSocket channel uses in `DevWorkspaceApiService`. This eliminates separate watch
infrastructure and lets the listener call `stopWatching()` (clean unsubscribe) after injection.
`devworkspaceApi` is passed from the route handler and carries the user's OAuth token, so both
watch and polling use the user's identity throughout.

A shared `injectCredentials()` private helper removes duplication between the watch path and the
polling fallback.

**2. Polling fallback when the watch receives an ERROR event.**
If the watch reports an error (RBAC gap, network hiccup), a polling fallback kicks in: GET the
devworkspace via `devworkspaceApi.getByName()` every 10 s for up to 300 s. Injection happens on
`Running`; polling stops without injection on `Failed`, `Failing`, `Stopped`, `Stopping`, or
`Terminating`. The workspace key stays registered in `activeWatches` while polling is active so
duplicate start calls are still blocked.

### Screenshot/screencast of this PR

_N/A — backend-only behavioral fix, no UI changes._

### What issues does this PR fix or reference?
fixes https://redhat.atlassian.net/browse/CRW-11193

### Is it tested? How?

1. Deploy Dev Spaces with the dashboard image built from this PR.
2. Log in as a regular user (not cluster-admin).
3. Start a workspace from a factory URL or from the dashboard.
4. Once the workspace is Running, open a terminal.
[...]

PR #1677, opened 2026-09-18
fix(backend): fix kubeconfig injection race in slow-starting workspaces

olexii4:
### What does this PR do?

Fixes user identity loss in workspace terminals on slow-starting infrastructure (CNV/RHDP)
by running credential-injection polling in parallel with the Kubernetes Watch from the start,
not as a post-failure fallback.

#### Root Cause

`PostStartInjector` relied on a K8s Watch stream to detect when a workspace reaches `Running`
and inject the user's kubeconfig. Polling was only started after the watch failed or timed out
(300 s).

On slower infrastructure (CNV in RHDP), the watch stream can be silently dropped by an
intermediary proxy or load balancer. No ERROR event, no TCP RST — just silence. The watch sits
idle until the 300 s timeout fires.

The UDI container's `entrypoint.sh` waits only ~12 s for `~/.kube/config`. By the time the
polling fallback started, the terminal had already fallen back to the pod service account
identity.

PR #1629 fixed two narrower races (already-Running at watch start, 10 s → 2 s poll interval)
but did not address the silent watch drop because polling still only started after watch failure.

#### Fix

1. **Parallel polling from t=0** — GET every 2 s runs alongside the watch from the start.
   Three detection paths race: K8s Watch, parallel polling, and an immediate initial GET
   (LIST→STREAM race). Whichever detects `Running` first injects and tears down the others.

2. **Ownership guard (`isOwner()`)** — the `activeWatches` map value is the specific `cleanup`
   closure. Every callback checks function identity before acting, so a stale in-flight
   callback from an older invocation cannot tear down a newer one.

3. **Unified cleanup** — the separate `startPollingFallback` code path is removed. One
   `cleanupAll()` handles watch, poll interval, and overall timeout together.

4. **Entrypoint window warning** — if injection takes longer than 12 s (the UDI entrypoint.sh
   window), a warning is logged with the detection source and elapsed time.

5. **Overall timeout (10 min)** — replaces the two separate 300 s timeouts. Logs a diagnostic
   warning if neither watch nor polling detected a phase transition.

N/A — backend-only change.

fixes https://redhat.atlassian.net/browse/CRW-11193

### Is it tested? How?
1. Deploy Eclipse Che with the dashboard image from this PR
2. Start a workspace from a factory URL
3. Open a terminal and run `oc whoami` — should return the OpenShift user identity
4. Run `oc whoami -t` — should return a valid OAuth token

#### Release Notes

Fixed workspace terminals falling back to the pod service account identity on slow-starting
infrastructure by polling for workspace status in parallel with the Kubernetes Watch stream.

#### Docs PR

N/A


tolusha:
/che-ai-assistant ok-pr-review

Task completed.

tolusha:
@olexii4 
Is there is a way to reproduce the original issue?

packages/dashboard-backend/src/services/PostStartInjector.ts
tolusha, line 29: Do we really need POLL TIMEOUT ?
I think we have to wait until workspace is run or stopped/deleted without any timeout.

packages/dashboard-backend/src/services/PostStartInjector.ts
```diff
@@ -144,6 +146,38 @@ export class PostStartInjector {
           podmanApi,
         );
       });
+
+    // Immediate initial check: if the workspace reached RUNNING before the watch
+    // stream opened (narrow LIST→STREAM race window), the watch will never fire for
+    // the transition.  A single GET right after registering the watch catches this.
```
tolusha: **The fix described in the PR body is not here.**

This single GET at t=0 closes the LIST->STREAM window, which is real. It does not cover the root cause the description names: a watch stream silently dropped by an intermediary after it opens.

In that scenario this GET returns `Starting`, no ERROR event ever arrives, and nothing detects `Running` until `INJECTION_TIMEOUT_MS` fires at 300 s. A workspace ready at t=30 s is injected at t=300 s, long after the entrypoint window closed.

Two options that may be cheaper than parallel polling:

- add `timeoutSeconds` to the watch request and re-establish - a watch that recycles every 60 s cannot sit dead for 300 s;
- use `resourceVersion: '0'` instead of `''` at `:134`, so the watch replays current state and the LIST->STREAM gap closes inside the protocol, without a second code path.

Also, a small wording point: the comment says "right after registering the watch", but `:133` only *starts* registering it. The watch is not established at this point, which is what the next comment is about.

packages/dashboard-backend/src/services/PostStartInjector.ts
```diff
@@ -144,6 +146,38 @@ export class PostStartInjector {
...
+        if (!PostStartInjector.activeWatches.has(key)) {
+          return;
+        }
+        const phase = dw.status?.phase;
+        const devworkspaceId = dw.status?.devworkspaceId;
+        if (phase === DevWorkspaceStatus.RUNNING && devworkspaceId) {
+          logger.info(
+            `PostStartInjector: ${key} already Running at watch start — injecting immediately`,
+          );
+          cleanupWithTimeout();
```
tolusha: **`cleanupWithTimeout()` here can leak the watch stream.**

`cleanup()` calls `devworkspaceApi.stopWatching()`, which is `this.stopWatch?.()`. In `devWorkspaceApi.ts:207`, `this.stopWatch` is assigned only *after* `await this.customObjectWatch.watch(...)` resolves.

This `getByName` is a plain REST round trip started in the same tick as `watchInNamespace()` at `:133`. When the workspace is already `Running` - the exact case this block was added for - the GET can resolve first. `stopWatching()` then finds `stopWatch` undefined and silently does nothing. `watchInNamespace()` afterwards installs a live namespace-wide stream with its timeout already cleared and its key already deleted, so nothing can abort it.

Two consequences:

1. A streaming connection to the API server leaks per occurrence.
2. The `listener` closure is still attached and does not check `activeWatches`, so the next `MODIFIED` event re-enters the `RUNNING` branch at `:121` and injects a second time - a second exec into a live container writing `~/.kube/config`.

The stale `activeWatches` entry is the worst part: `:60` will then skip every future `watchAndInject()` for this key until the pod restarts, so a user restarting that workspace gets no injection at all.

This race did not exist before - `cleanup()` was previously only reachable from the listener (which implies the watch is established) or from the 300 s timeout.

Would it be possible to have `watchInNamespace()` resolve only once the watch is abortable, or return the abort handle?

tolusha (commented):
## Automated review (`ok-pr-review`)

Ran summary, review, deep-review and impact. Verdicts: 🔄 **Request Changes** / ⚠️ Design needs refinement / ⚠️ Concerns.

---

### 1. This branch contains no change relative to `main`

Both changed files at `8e09edc3` are byte-identical to `main`:

| File | blob @ `main` | blob @ `8e09edc3` |
|---|---|---|
| `PostStartInjector.ts` | `057df9528b...` | `057df9528b...` |
| `__tests__/PostStartInjector.spec.ts` | `255aefb740...` | `255aefb740...` |

This content was already merged by #1629 (squash merge `1cc7697a`, 2026-07-08). This PR reuses the same branch `CRW-11193` at the same commit. Because #1629 was squash-merged, `8e09edc3` is not an ancestor of `main` (`compare main...8e09edc3` -> `diverged, ahead_by: 1, behind_by: 37`), so GitHub re-renders the already-merged change as a fresh diff.

Merging would change nothing. Does it make sense to either close this as a duplicate of #1629, or force-push the intended work onto the branch?

### 2. The PR description does not match the diff

The body states:

> Run polling (GET every 2 s) **in parallel with the watch from the start**, not as a post-failure fallback.
> The separate `startPollingFallback` code path is removed.

Neither is true here. `startPollingFallback` is fully intact at `PostStartInjector.ts:183`, and polling still starts only on watch error (`:93`), watch rejection (`:141`), or the 300 s timeout (`:75`). The commit message on `8e09edc3` describes the actual change accurately - it looks like the body was written against work that was not pushed.

This matters because the root cause the description names - a silently dropped watch stream - is not addressed by anything in this branch.

### What was found in the code itself

The code is on `main` already, so these apply to current production behaviour rather than to the merge decision. Details are inline.

**Two resource leaks introduced by the initial-GET block.** Both end with a stale entry left in `activeWatches`, which then blocks *all* future injection for that workspace key until the pod restarts (`:60`). That is the most operationally serious consequence.

**The stated root cause is still open.** A watch that dies silently produces no ERROR event, so nothing detects `Running` until `INJECTION_TIMEOUT_MS` fires at 300 s - long after the UDI entrypoint window closes. Two cheaper options than parallel polling are worth considering: an explicit `timeoutSeconds` on the watch with re-establishment, or `resourceVersion: '0'` so the watch replays current state and closes the LIST->STREAM gap inside the protocol.

**Observability.** This PR exists because identity loss was discovered by reading `KubedockPodmanTest` CI logs. A failed `injectKubeConfig()` still produces only an `error` log line - no counter, no DevWorkspace event, no status condition. `injectCredentials()` already computes `elapsedMs` at `:262-263` and discards it into a log string. A failure event would make the silent service-account fallback visible via `kubectl describe dw`.

**Poll rate.** 10 s -> 2 s is right for the user-facing problem. It also takes the fallback from 30 to 150 GETs per workspace, with a fixed period and no jitter, so a correlated failure (operator rollout, API-server restart) puts every affected request into a synchronised loop against the API server that just recovered. 2 s with +/-25% jitter keeps the latency win.

**Tunables.** `INJECTION_TIMEOUT_MS`, `POLL_INTERVAL_MS` and `POLL_TIMEOUT_MS` are all compile-time constants. This PR exists because one of them needed changing, which cost a full release cycle - and #1606, #1620, #1629 and this PR all target CRW-11193. Env-var overrides with the current values as defaults would make the next iteration cheap.

### Open question from @tolusha

> Do we really need POLL TIMEOUT? I think we have to wait until workspace is run or stopped/deleted without any timeout.

Worth one caution if that change is made: with the timeout gone, `isTerminalPhase` becomes the only exit, so a DevWorkspace deleted mid-poll would keep the interval alive forever - `getByName` rejects with 404 and the `.catch` at `:248-250` only logs and retries. An unbounded poll should treat 404 as a stop condition.

### Also worth answering

> Is there a way to reproduce the original issue?

The current test plan is manual (`oc whoami`), which cannot reproduce a race. The `KubedockPodmanTest` CI log cited in #1629's description is probably the best available evidence - linking it here would help the next person.

### What was done well

- The commit message on `8e09edc3` is exemplary: it names both races, explains why each matters, cites the CI evidence, and links the Jira.
- The `POLL_INTERVAL_MS` comment explains *why* 2 s by tying it to an external constraint, rather than restating the literal.
- `injectCredentials()` uses separate try/catch per step, so a podman failure cannot discard a successful kubeconfig injection.
- The double-injection guard at `:158` is correct for the case it targets - both the listener and the initial GET run their check and cleanup synchronously, so the event loop cannot interleave them.
- The comments at `:200-203` and in `routes/api/devworkspaces.ts:71-74` pre-empt reviewer questions about `devworkspaceApi` instance lifetime with actual reasons.

<sub>Generated by <code>ok-pr-review</code> (summary + review + deep-review + impact).</sub>

packages/dashboard-backend/src/services/PostStartInjector.ts
```diff
@@ -144,6 +146,38 @@ export class PostStartInjector {
...
+
+    // Immediate initial check: if the workspace reached RUNNING before the watch
+    // stream opened (narrow LIST→STREAM race window), the watch will never fire for
+    // the transition.  A single GET right after registering the watch catches this.
+    devworkspaceApi
+      .getByName(namespace, workspaceName)
+      .then(async dw => {
+        // Guard: if the watch listener already handled RUNNING (and called
+        // cleanupWithTimeout), the key is gone — avoid double injection.
+        if (!PostStartInjector.activeWatches.has(key)) {
```
tolusha: **This guard asks a question the map cannot answer.**

`activeWatches` is used for three different things: a re-entrancy guard (`:60`), a cleanup registry (`:70`, `:204`), and - new here - an ownership token meaning "the watch still owns this workspace".

It does not carry that third meaning. `startPollingFallback()` re-registers the same key at `:204` with its own cleanup closure. So if the watch errors before this GET resolves, `has(key)` is `true` but the owner is the poll loop.

This block then proceeds and calls `cleanupWithTimeout()`, which invokes the **watch's** `cleanup` captured at `:65-68` - not the polling `cleanup` at `:195-198`. The poll loop never gets `cancelled = true` and never gets `clearInterval`, so it keeps running for up to `POLL_TIMEOUT_MS` and injects a second time on its next tick.

Making ownership explicit would fix this class of bug rather than this instance:

```ts
type Owner = { kind: 'watch' | 'poll'; cancel: () => void };
private static activeWatches = new Map<string, Owner>();
```

Then this guard becomes `activeWatches.get(key)?.kind === 'watch'`, and all teardown goes through `activeWatches.get(key)?.cancel()` instead of a lexically captured closure. Worth applying the same check to the listener at `:121` and the poll tick at `:235`, so no path can inject after teardown.

packages/dashboard-backend/src/services/PostStartInjector.ts
```diff
@@ -17,7 +17,9 @@ import { MessageListener } from '@/services/types/Observer';
 import { logger } from '@/utils/logger';
 
 const INJECTION_TIMEOUT_MS = 300000;
-const POLL_INTERVAL_MS = 10000;
+// 2 s gives ~6 poll attempts within the ~12 s window that the UDI entrypoint.sh
+// waits for ~/.kube/config before giving up and falling back to the pod SA token.
+const POLL_INTERVAL_MS = 2000;
```
tolusha, line 24: Two things on this constant.

**Jitter.** 10 s -> 2 s is right for the user-facing problem, but it also takes the fallback from 30 to 150 GETs per workspace over the 300 s window, on a fixed period. When many watches fail at once (operator rollout, API-server restart) every affected request enters a synchronised 2 s loop against the API server that just recovered. 2 s with +/-25% jitter keeps the latency benefit and breaks the synchronisation.

**Externalising.** This value, `INJECTION_TIMEOUT_MS` and `POLL_TIMEOUT_MS` are all compile-time constants. This PR exists because one of them needed to change, and that cost a code edit, a review, an image build and a release - and #1606, #1620, #1629 and this PR all target CRW-11193. An env-var override with the current value as the default would make the next iteration cheap, and would let fast clusters keep a slower interval instead of paying 5x the fallback traffic for no benefit.

Also, the class doc at `:45` still reads "Falls back to polling (GET every 10 s, up to 300 s)":

```suggestion
const POLL_INTERVAL_MS = 2000;
```
(no change needed on this line - just flagging `:45` needs the same update)

packages/dashboard-backend/src/services/PostStartInjector.ts
```diff
@@ -17,7 +17,9 @@ import { MessageListener } from '@/services/types/Observer';
 import { logger } from '@/utils/logger';
 
 const INJECTION_TIMEOUT_MS = 300000;
-const POLL_INTERVAL_MS = 10000;
+// 2 s gives ~6 poll attempts within the ~12 s window that the UDI entrypoint.sh
```
tolusha, line 22: This is the good kind of magic-number comment - it explains *why* 2 s rather than restating the literal.

One gap: the ~12 s figure is a constant in the UDI image, in a different repository. There is no shared constant, no integration test, and no CI job that would fail if it changed. If UDI's wait drops to 5 s, this silently stops working and the only symptom is intermittent identity loss.

Could we name the source here - repo and file - so the number is auditable later?

packages/dashboard-backend/src/services/__tests__/PostStartInjector.spec.ts
```diff
@@ -45,7 +45,9 @@ describe('PostStartInjector', () => {
         return Promise.resolve();
       }),
       stopWatching: jest.fn(),
-      getByName: jest.fn(),
+      // Default: return 'Starting' so the initial check does not trigger injection
+      // in tests that do not explicitly override getByName.
+      getByName: jest.fn().mockResolvedValue({ status: { phase: 'Starting' } }),
```
tolusha, line 49: This default is a sensible way to keep the existing tests green.

Worth noting a limitation of the double just above it though: `watchInNamespace` at `:43-46` captures the listener **synchronously** and resolves immediately, and `stopWatching` at `:47` is inert.

The real implementation captures nothing synchronously and only becomes abortable after an awaited network call. So the mock guarantees the one property production does not - that `stopWatching()` always works - which is exactly the property the new initial-GET block depends on. No test in this file can fail if that teardown leaks.

If `watchInNamespace` returned a controllable deferred, and `stopWatching` actually detached the captured listener, the leak would become observable in a test.

packages/dashboard-backend/src/services/__tests__/PostStartInjector.spec.ts
```diff
@@ -194,9 +196,45 @@ describe('PostStartInjector', () => {
...
+      status: { phase: 'Running', devworkspaceId: 'ws-already-running' },
+    });
+
+    invoke();
+    // Flush the initial getByName() promise chain
+    await Promise.resolve().then(() => Promise.resolve());
+
+    expect(kubeConfigApi.injectKubeConfig).toHaveBeenCalledWith(namespace, 'ws-already-running');
+    expect(podmanApi.podmanLogin).toHaveBeenCalledWith(namespace, 'ws-already-running');
+    expect((PostStartInjector as any).activeWatches.has(key)).toBe(false);
```
tolusha, line 220: This asserts the bookkeeping, but not the effect.

`activeWatches.has(key) === false` tells us the map entry was removed. It does not tell us the watch was torn down - and this test exercises the exact path where `stopWatching()` can be a no-op, because the initial GET may resolve before `watchInNamespace()` has assigned its abort handle.

Would it be worth adding `expect(devworkspaceApi.stopWatching).toHaveBeenCalled()` here? It is what the sibling tests at `:137` and `:175` do, and its absence in the one new test that needs it most is easy to miss.

packages/dashboard-backend/src/services/__tests__/PostStartInjector.spec.ts
```diff
@@ -210,16 +248,18 @@ describe('PostStartInjector', () => {
...
   });
 
-  test('injects credentials via polling fallback after 60 s watch timeout', async () => {
-    (devworkspaceApi.getByName as jest.Mock).mockResolvedValue({
-      status: { phase: 'Running', devworkspaceId: 'ws-timeout-id' },
-    });
+  test('injects credentials via polling fallback after watch timeout', async () => {
+    // First call = initial check → Starting (no injection).
+    // Subsequent calls (polling) → Running.
+    (devworkspaceApi.getByName as jest.Mock)
```
tolusha: These `mockResolvedValueOnce(...).mockResolvedValue(...)` chains assert against the number and order of `getByName` calls, which is an implementation detail rather than behaviour. The same pattern is at `:292`, `:335` and `:352`.

The concrete cost: the parallel-polling change this PR's description promises adds calls to that sequence. All four tests would break even though the observable behaviour - credentials injected once, promptly - is unchanged or better.

A phase-driven mock decouples them:

```ts
let currentPhase = 'Starting';
getByName: jest.fn(() =>
  Promise.resolve({ status: { phase: currentPhase, devworkspaceId: 'ws-id' } }),
),
```

Tests then set `currentPhase = 'Running'` at the point they care about, and stay green regardless of how many GETs the implementation makes.

packages/dashboard-backend/src/services/__tests__/PostStartInjector.spec.ts
```diff
@@ -287,18 +329,38 @@ describe('PostStartInjector', () => {
       expect((PostStartInjector as any).activeWatches.has(key)).toBe(false);
     });
 
+    test('poll interval is 2 s (catches RUNNING before entrypoint.sh gives up)', async () => {
```
tolusha: This test does not actually verify the interval.

It advances timers by exactly 2000 ms once and asserts injection happened. It would pass identically with `POLL_INTERVAL_MS` set to 500, 1000 or 2000 - any value that divides into 2000. As written it is close to a duplicate of `injects on Running phase during polling` at `:290` with a different workspace id.

The negative half is what pins the value:

```suggestion
    test('poll interval is 2 s (catches RUNNING before entrypoint.sh gives up)', async () => {
      // The UDI entrypoint.sh waits ~12 s; 2 s poll gives ~6 attempts in that window.
      // First call = initial check → Starting. Poll at T+2 s → Running.
      (devworkspaceApi.getByName as jest.Mock)
        .mockResolvedValueOnce({ status: { phase: 'Starting' } })
        .mockResolvedValue({ status: { phase: 'Running', devworkspaceId: 'ws-fast-id' } });

      await triggerErrorAndFlush();

      // Nothing before the interval elapses
      jest.advanceTimersByTime(1999);
      await Promise.resolve().then(() => Promise.resolve());
      expect(kubeConfigApi.injectKubeConfig).not.toHaveBeenCalled();

      // ...and injection on the first 2 s tick
      jest.advanceTimersByTime(1);
      await Promise.resolve().then(() => Promise.resolve());

      expect(kubeConfigApi.injectKubeConfig).toHaveBeenCalledWith(namespace, 'ws-fast-id');
    });
```

Unrelated but nearby: `await Promise.resolve().then(() => Promise.resolve())` appears 9 times in this file and drains exactly two microtask ticks. It works only because the chains happen to be that deep - one extra `await` inside `injectCredentials()` would break several tests silently. A single `flushPromises()` helper would be depth-independent.

packages/dashboard-backend/src/services/__tests__/PostStartInjector.spec.ts
```diff
@@ -194,9 +196,45 @@ describe('PostStartInjector', () => {
...
+    // Now flush the initial check promise — key is already gone
+    await Promise.resolve().then(() => Promise.resolve());
+
+    expect(kubeConfigApi.injectKubeConfig).toHaveBeenCalledTimes(1);
+  });
+
   // ── timeout ───────────────────────────────────────────────────────────────
 
   test('stops watch and starts polling fallback on 300 s timeout', () => {
+    // Return 'Starting' so the initial check does not trigger injection.
```
tolusha: Two small things here.

This override is now redundant - `beforeEach` at `:50` already defaults `getByName` to `Starting`.

More importantly, the test is not `async`, but `invoke()` now starts an unawaited `getByName()` chain. It settles after the test body returns, inside a later test's window. It happens to be harmless because `beforeEach` replaces `activeWatches` with a fresh Map, so the `:158` guard short-circuits - but the test is passing by accident of cleanup ordering. Making it `async` and flushing once would make that intentional.

tolusha (changes requested):

packages/dashboard-backend/src/services/PostStartInjector.ts
olexii4: Agreed. Removed `POLL_TIMEOUT_MS` entirely. Polling now runs until one of three things happens:
- `Running` detected → inject and clean up
- Terminal phase detected → clean up
- Overall 10 min timeout → clean up with diagnostic warning

The overall timeout is a safety net against leaked resources, not a behavioral deadline. The polling itself is unbounded within that window.

One note on your concern about 404: the `.catch` on the poll GET logs and retries, which handles transient network errors. A persistent 404 (workspace deleted) means the DevWorkspace object is gone, so no `Running` will ever arrive and the overall timeout will clean up. If we want faster exit on 404 specifically, we can add that as a follow-up — but it would only save the tail of the 10 min window.

tolusha: I still have couple of question
1. Where does `ENTRYPOINT_WINDOW_MS` come from?
2. What if workspace starts longer than 10 min?

olexii4: @tolusha

1. `ENTRYPOINT_WINDOW_MS` comes from the UDI image — `devspaces-images/devspaces-udi/etc/entrypoint.sh` has a `KUBECONFIG_WAIT_TIMEOUT` variable set to ~12 s. If `~/.kube/config` is not present within that window, the entrypoint falls back to the pod service account token. The constant here mirrors that value so the dashboard can log a warning when injection is likely too late.

2. If a workspace takes longer than 10 min to start, the overall timeout fires and cleans up the watch + poll resources. The workspace itself is not affected — it continues starting normally. The only consequence is that kubeconfig injection will not happen automatically. The user would need to restart the workspace (which triggers a new `watchAndInject` call) or the kubeconfig would need to be injected manually.

In practice, a 10 min startup on any supported infrastructure means something is stuck (image pull timeout, scheduling failure, etc.) — the workspace will likely hit its own DevWorkspace timeout before ours.

packages/dashboard-backend/src/services/PostStartInjector.ts
olexii4: Fixed. Parallel polling from t=0 is now implemented. The watch ERROR path no longer starts a separate polling fallback — it calls `devworkspaceApi.stopWatching()` and the existing poll loop continues.

Regarding your two alternatives:

- `timeoutSeconds` with re-establishment would catch dead streams. The trade-off is that `watchInNamespace()` currently handles one watch per namespace — recycling would need a reconnect loop. The parallel poll gives the same detection coverage without touching the watch infrastructure.
- `resourceVersion: '0'` would replay current state and close the LIST→STREAM gap inside the protocol. However, using `'0'` returns the latest cached state from the API server, which may not include the most recent transition on a multi-node etcd cluster. The initial GET + parallel poll covers the same case more reliably.

Neither rules out the other — we can layer `timeoutSeconds` on top of parallel polling in a follow-up.

packages/dashboard-backend/src/services/PostStartInjector.ts
olexii4: Fixed. The entire approach changed:

1. The `isOwner()` guard replaces `activeWatches.has(key)`. It checks function identity (`activeWatches.get(key) === cleanup`), so a stale closure from an older invocation is a no-op even if the key exists.
2. `cleanup()` itself checks `isOwner()` before doing anything. If a newer invocation has replaced the map entry, the old cleanup silently returns.
3. There is no separate initial-GET-then-teardown path anymore. The initial GET goes through `handleRunning()`, which calls `cleanupAll()`, which calls `cleanup()` — same path as the watch listener and the poll.

The double-injection scenario you described (stale `activeWatches` entry blocking future injection) cannot happen because `cleanup()` is a no-op when it is not the current owner.

packages/dashboard-backend/src/services/PostStartInjector.ts
olexii4: Fixed, though with a simpler approach than the `Owner` type you suggested. The map value is the specific `cleanup` closure. Every callback (`handleRunning`, `handleTerminal`, poll tick, initial check) calls `isOwner()` which does `activeWatches.get(key) === cleanup`. This is equivalent to your `kind` discriminant — the closure identity *is* the ownership token.

The `startPollingFallback` code path is removed entirely, so there is no dual-owner problem. Watch, poll, and initial GET all share the same `cleanup` closure registered at invocation time.

The ownership guard test confirms this:

```
'stale poll from old invocation does not tear down new invocation'
```

Uses a deferred promise to simulate the race: first invocation starts, terminal phase tears it down, second invocation starts, then the first invocation's initial-check promise resolves with `Running` — it must be a no-op.

packages/dashboard-backend/src/services/PostStartInjector.ts
olexii4: Both are good points. Will address in a follow-up:
- Jitter on `POLL_INTERVAL_MS` (randomize each tick within ±25%)
- Env-var overrides for `OVERALL_TIMEOUT_MS`, `POLL_INTERVAL_MS`, `ENTRYPOINT_WINDOW_MS`

Keeping this PR focused on the structural fix (parallel polling + ownership guard).

packages/dashboard-backend/src/services/PostStartInjector.ts
olexii4: Thanks. The constant lives in `devspaces-images/devspaces-udi/etc/entrypoint.sh` — the `KUBECONFIG_WAIT_TIMEOUT` variable. Will add a comment referencing the source file.

packages/dashboard-backend/src/services/__tests__/PostStartInjector.spec.ts
olexii4: Acknowledged. The current mock is a simplification. The ownership guard test works around this by using two separate `devworkspaceApi` instances — the second invocation gets its own mock with its own `stopWatching`, so the test verifies that the stale callback does not call either one.

A more realistic mock with a controllable deferred for `watchInNamespace` and a proper `stopWatching` that detaches the listener would be a good improvement. Will track as a follow-up test improvement.

packages/dashboard-backend/src/services/__tests__/PostStartInjector.spec.ts
olexii4: Added. The `injects and logs unsubscribe when watch detects Running` test now asserts `stopWatching` was called. The overall timeout tests also assert `stopWatching`.

packages/dashboard-backend/src/services/__tests__/PostStartInjector.spec.ts
olexii4: Partially addressed. The new tests use `mockResolvedValueOnce` chains but the structure is simpler because there is no separate polling fallback to set up. The `injects via poll when watch is silent` test explicitly labels each mock return (initial check, poll #1, poll #2) which makes the chain readable.

A `let currentPhase` approach would decouple further — can refactor in a follow-up if the current chains become brittle again.

packages/dashboard-backend/src/services/__tests__/PostStartInjector.spec.ts
olexii4: The new `polls every 2 s alongside the watch` test counts `getByName` calls at 2 s intervals:
- At t=2s: 2 calls (initial check + first poll)
- At t=4s: 3 calls

This verifies the interval is 2 s and not shorter. The negative half (nothing at t=1999ms) is not explicitly tested but could be added.

packages/dashboard-backend/src/services/__tests__/PostStartInjector.spec.ts
olexii4: Fixed. The redundant override is removed. All tests that call `invoke()` are now `async` and flush microtasks properly.

olexii4:
@tolusha the root cause is a silently dropped K8s Watch stream — no ERROR event reaches the dashboard. This happens when an intermediary proxy or LB terminates the long-lived HTTP/2 stream without sending a TCP RST.

Reproduction on CNV/RHDP:

1. Start a workspace on a slow cluster where startup takes >30 s
2. The watch stream passes through an intermediary that has an idle timeout shorter than the workspace startup time
3. The watch sits idle, no `MODIFIED` event arrives, and injection waits for the 300 s timeout
4. `oc whoami` in the terminal returns the pod SA identity instead of the user

The `KubedockPodmanTest` CI failures on RHDP were the original evidence — the pod SA token caused podman login to fail because the SA lacks the registry credentials the user's OAuth token carries.

With parallel polling, even if the watch dies silently, the 2 s poll catches `Running` within ~4 s and injects before the 12 s entrypoint window closes.

olexii4:
@tolusha thanks for the thorough review. The two new commits implement the changes described in the PR body: parallel polling from t=0, `isOwner()` ownership guard, unified cleanup, and `startPollingFallback` removal.

Replied to all inline comments. Items deferred to follow-up:

| Item | Reason |
|---|---|
| Poll jitter (±25%) | Separate concern, simple change |
| Env-var overrides for timeouts | Separate concern, needs documentation |
| `timeoutSeconds` on watch with re-establishment | Needs changes to `watchInNamespace()` API |
| `resourceVersion: '0'` evaluation | Needs testing on multi-node etcd |
| UDI source reference comment | Small, can go in next touch |
| Phase-driven mock refactor | Test quality improvement, no behavior change |
| Controllable deferred mock for `watchInNamespace` | Test quality improvement |
| 404 as explicit stop condition in poll | Edge case, overall timeout handles it |

packages/dashboard-backend/src/services/PostStartInjector.ts
tolusha: 1. I can't find `KUBECONFIG_WAIT_TIMEOUT` in the UDI entrypoint. Could you share a direct link to it, please?
2. That's exactly what we're trying to avoid. We don't need any timeouts as long as the workspace is in the `starting` state. At the very least, we should make them extremely long, e.g. 1 hour.

tolusha:
There are 2 questions, why do we need:
1.  KUBECONFIG_WAIT_TIMEOUT?
2. Overall timeout (10m) ?

packages/dashboard-backend/src/services/PostStartInjector.ts
olexii4: @tolusha you're right, there is no `KUBECONFIG_WAIT_TIMEOUT` variable in the UDI entrypoint — my previous comment was incorrect.

The actual variable is `KUBEDOCK_TIMEOUT` (default 10 s), and it only applies when `KUBEDOCK_ENABLED=true`. The relevant section in `devspaces-udi/build/udi/etc/entrypoint.sh`:

```sh
KUBEDOCK_TIMEOUT=${KUBEDOCK_TIMEOUT:-10}
until [ -f $KUBECONFIG ]; do
    if (( SECONDS > KUBEDOCK_TIMEOUT )); then
        break
    fi
    echo "Kubeconfig doesn't exist yet. Waiting..."
    sleep 1
done
```

This waits for `$KUBECONFIG` to appear before starting kubedock. It is not a general kubeconfig injection window — it only gates kubedock startup.

The `ENTRYPOINT_WINDOW_MS` constant has been removed in the latest commit (`71ecc64`). The dashboard no longer tracks or warns about any entrypoint timing window.

